PR #4, opened 2021-01-10
Cease using SSE4.2 instructions for CRC

agrover:
Prev code caused mkfs to fail to run on CPUs without SSE4.2.

(This includes my colo box, where I'm trying to get self-hosted CI working from.)

Remove hand-coded CRC calc in favor of implementation from zlib. zlib
is already an implicit dependency (for openssl) so no additional
packages required.


agrover:
didn't work out.

PR #6, opened 2021-01-14
Fix mkfs check for existing ScoutFS superblock

agrover:
We were checking for the wrong magic value.

We now need to use -f when running mkfs in run-tests for things to work.


zabbo:
Nice, thanks for pushing this.  I added these exact changes to the log merging branch yesterday :).

PR #7, opened 2021-01-14
Filesystem version instead of format hash check

agrover:
Instead of hashing headers, define an interop version. Do not mount
superblocks that have a different version, either higher or lower.

Since this is pretty much the same as the format hash except it's a
constant, minimal code changes are needed.

Initial dev version is 0, with the intent that version will be bumped to
1 immediately prior to tagging initial release version.

Update README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,15 +31,9 @@ functionality hasn't been implemented.  It's appropriate for early
 adopters and interested developers, not for production use.
 
 In that vein, expect significant incompatible changes to both the format
-of network messages and persistent structures.  To avoid mistakes the
-implementation currently calculates a hash of the format and ioctl
-header files in the source tree.  The kernel module will refuse to mount
-a volume created by userspace utilities with a mismatched hash, and it
-will refuse to connect to a remote node with a mismatched hash.  This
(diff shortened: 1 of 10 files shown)


agrover:
I tested this manually, but didn't see a good way to add an automated test.

I think once there are actually more than one version in existence we will be in a better position to add a test (and indeed we will want to make sure we do.)

zabbo (changes requested):
We had talked about having a mount advertise its version somehow.  Now that I think about it, it doesn't need to be in sysfs 'cause it's baked into the built module.  We can add a elf note for the version just like we do for the git-describe string.  Then you can also see it in the module .ko with modinfo.  Add another inline asm to the top of scoutfs_module_init() for the interop_version and make sure they both appear in /sys/module/scoutfs/notes/.  I agree that we can keep this initial commit small, as we increase the version to 1 we can add triggers to use bad versions for tests to exercise.  So with that elf note and the utils/ makefile fixes I think we're good

agrover:
* put back FMTIOC stuff for make dist
* add interop version to modinfo and notes
* remove hash-related comment in utils/Makefile

PR #10, opened 2021-01-18
Remove item accounting

agrover:
Remove item accounting.


kmod/src/scoutfs_trace.h
```diff
@@ -426,11 +425,15 @@ TRACE_EVENT(scoutfs_trans_write_func,
 
 TRACE_EVENT(scoutfs_release_trans,
 	TP_PROTO(struct super_block *sb, void *rsv, unsigned int rsv_holders,
-		 struct scoutfs_item_count *res,
-		 struct scoutfs_item_count *act, unsigned int tri_holders,
+		 int reserved_items,
```
zabbo, line 428: Oh, if we're removing the item accounting, we're not just removing the arguments to hold and release.  We'd be removing all the accounting in the transactions too.  That's the book keeping to remove :).  So no more reserved/actual fields.

zabbo (changes requested):
If we're going to do this, let's do it -- I'm cool with either dropping the removal or going the rest of the way.

agrover:
Like this?

Is there more??

zabbo:
That was it!  There's still refinement to do -- the "reservation" isn't anymore, it's just a per-task count of holders so that nested holds don't block waiting for their caller's hold to release.  We wouldn't need to do that with a full allocated struct, we could just use the task_struct journal_info pointer itself, but we can leave all that for the day when we want to make the transaction holds less expensive.  And we might need it again if we have to do reservations to properly address the premature enospc case over in the metadata allocators.

PR #16, opened 2021-01-29
Inode Deletion

agrover:
Track open inodes in a counted bloom filter, which is saved to a
non-counted bloom filter metadata block when the transaction closes.


agrover:
@zabbo Here's a draft PR for part 1, if you can do a prelim look-see?

zabbo (changes requested):
Looking pretty good so far!  I just did a quick scan in github, we can iterate more aggressively as it gets further along.

kmod/src/forest.c
```diff
@@ -662,7 +687,131 @@ void scoutfs_forest_destroy(struct super_block *sb)
...
+void scoutfs_forest_oino_opened(struct super_block *sb, struct inode *inode)
+{
+	struct scoutfs_inode_info *si = SCOUTFS_I(inode);
+	DECLARE_FOREST_INFO(sb, finf);
+	u64 hash;
+	int val;
+	int i;
+
+	val = atomic_inc_return(&finf->oino_open_count);
+	WARN_ON_ONCE(val <= 0);
```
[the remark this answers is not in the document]
agrover: Nah this is fine. I had a misunderstanding of atomic_inc_return semantics.

kmod/src/forest.c
```diff
@@ -662,7 +687,131 @@ void scoutfs_forest_destroy(struct super_block *sb)
...
+
+	bb = new_bl->data;
+
+	/* use different magic? */
+	bb->hdr.magic = cpu_to_le32(SCOUTFS_BLOCK_MAGIC_BLOOM);
+	bb->hdr.fsid = SCOUTFS_SB(sb)->super.hdr.fsid;
+	bb->hdr.blkno = cpu_to_le64(blkno);
+	bb->hdr.seq = cpu_to_le64(trans_seq);
+
+	ref = &finf->our_log.oino_bloom_ref;
```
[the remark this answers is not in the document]
agrover: ok, yes.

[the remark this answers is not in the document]
agrover: The first thing I tried -- call scoutfs_free_meta(blkno) if old oino_bloom_ref was nonzero -- did not work: server error emptying freed (-EINVAL) because of overlap detected in scoutfs_ext_insert().

If this isn't working because the block is from a previous transaction, how does one free said block?

agrover:
* Fixed easy stuff, still broken/not freeing prev oino blk ref.

kmod/src/forest.c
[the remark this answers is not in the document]
zabbo: So what's happening here is that the client is creating a new dirty version of the oino bloom block to write and be applied by the server as the new version referenced by the log_trees item that everyone will read and use.  When the server applies the client's transaction it will be overwriting the existing ref, so part of applying the transaction needs to also free the old blkno.   That's what freeing the old non-zero ref blkno in the client does as it dirties the new oin bloom block -- it adds the old ref blkno to the freed alloc list block that will be given to the server as the transaction is applied.  Dunno why you'd be seeing double frees (what the overlap is finding), we'd have to see the code to tell for sure.  It'd want to look like the freeing of the old forest item bloom block in _bloom_set_bits().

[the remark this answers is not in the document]
agrover: duh, I'm freeing it twice for some reason is why :facepalm:

kmod/src/forest.c
```diff
@@ -662,7 +687,131 @@ void scoutfs_forest_destroy(struct super_block *sb)
 
 	if (finf) {
 		scoutfs_block_put(sb, finf->srch_bl);
+		scoutfs_block_put(sb, finf->oino_bl);
+		vfree(finf->oino_counted_bloom);
```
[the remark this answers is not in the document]
agrover: Looks safe to me. `scoutfs_forest_destroy` is being called from `scoutfs_put_super`, which is called from `generic_shutdown_super`, which is called from `kill_block_super`, which then syncs and puts the blkdev. This is the stone-cold unmount/free path, there's nothing open at this point.

kmod/src/forest.c
```diff
@@ -620,12 +629,15 @@ void scoutfs_forest_get_btrees(struct super_block *sb,
 
 	lt->item_root = finf->our_log.item_root;
 	lt->bloom_ref = finf->our_log.bloom_ref;
+	lt->oino_bloom_ref = finf->our_log.oino_bloom_ref;
```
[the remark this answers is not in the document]
agrover: ok, took a stab at this, "this" being freeing the oino block from `reclaim_log_trees`.

agrover:
* freeing prev oino blkno works now
* free oino block from `reclaim_log_trees`
* Feeling good that open/releases will not occur after we vfree `counted_bloom`.

kmod/src/server.c
```diff
@@ -627,6 +627,9 @@ static int reclaim_log_trees(struct super_block *sb, u64 rid)
 					&lt.meta_avail) ?:
 	      alloc_move_empty(sb, &super->data_alloc, &lt.data_avail) ?:
 	      alloc_move_empty(sb, &super->data_alloc, &lt.data_freed);
+	if (lt.oino_bloom_ref.blkno && !ret)
+		ret = scoutfs_free_meta(sb, &server->alloc, &server->wri,
+					le64_to_cpu(lt.oino_bloom_ref.blkno));
```
zabbo, line 632: Looks right, but we'd want to zero that ref blkno (and maybe the whole ref including the seq?) to indicate that the block has been freed.  The log_trees item will still sit around for a while so that the item log tree and item bloom can be used by readers.

zabbo (approved):
Looking good for the first draft, we can clean up once we see it working with the reader side and start to see how much it all costs.

kmod/src/server.c
agrover: zero the ref, got it.

kmod/src/forest.c
```diff
@@ -801,23 +810,162 @@ int scoutfs_forest_oino_alloc(struct super_block *sb, u64 trans_seq)
...
+		} else {
+			bb = bl->data;
+
+			/* Use oldest block's seq */
+			first_bb->hdr.seq = cpu_to_le64(
+				min(le64_to_cpu(bb->hdr.seq),
+				    le64_to_cpu(first_bb->hdr.seq)));
+
+			for (i = 0; i < SCOUTFS_FOREST_BLOOM_ARRAY_LEN; i++) {
+				first_bb->bits[i] |= bb->bits[i];
```
zabbo, line 910: bitmap_or(priv_bb->bits, priv_bb->bits, bb->bits) and then we don't need the _ARRAY_LEN define either. (which could have been ARRAY_SIZE(bb->bits) also, I think.

kmod/src/forest.c
```diff
@@ -801,23 +810,162 @@ int scoutfs_forest_oino_alloc(struct super_block *sb, u64 trans_seq)
...
+			}
+
+			scoutfs_block_put(sb, bl);
+		}
+	}
+
+	if (first_bb) {
+		ret = scoutfs_scan_orphans(sb, first_bb);
+		if (ret > 0) {
+			schedule_work(&finf->oino_work.work);
```
zabbo, line 920: Hmm, I don't think cancel_delayed_work_sync() can necessarily wait for normal scheduled work?  there's very confusing object layering going on in there.  I'd use schedule_delayed_work(, 0) just to be safe.

kmod/src/forest.c
```diff
@@ -801,23 +810,162 @@ int scoutfs_forest_oino_alloc(struct super_block *sb, u64 trans_seq)
...
+	if (orphan_seq > le64_to_cpu(bb->hdr.seq)) {
+		scoutfs_inc_counter(sb, forest_oino_deletable_false_1);
+		return false;
+	}
+
+	/* ino not deletable until bloom says it's closed everywhere */
+	hash = scoutfs_hash64(&le_ino, sizeof(le_ino));
+	for (i = 0; i < SCOUTFS_FOREST_BLOOM_NRS; i++) {
+		u32 idx = (u32)hash % SCOUTFS_FOREST_BLOOM_BITS;
+		bloom_hits += !!test_bit_le(idx, bb->bits);
```
zabbo, line 959: Just break if test_bit_le() is false, then use i == _NRS to detect if they all passed.  The typical case is that unlinked files won't be open so we'll save some time.

kmod/src/format.h
```diff
@@ -458,6 +458,9 @@ struct scoutfs_bloom_block {
  * positives.
  */
 #define SCOUTFS_FOREST_BLOOM_NRS		3
+#define SCOUTFS_FOREST_BLOOM_ARRAY_LEN \
```
zabbo, line 461: (re: bitmap_or() use above, I bet we can get rid of this)

kmod/src/inode.c
```diff
@@ -1566,12 +1565,15 @@ int scoutfs_drop_inode(struct inode *inode)
...
-int scoutfs_scan_orphans(struct super_block *sb)
+int scoutfs_scan_orphans(struct super_block *sb,
+			 struct scoutfs_bloom_block *bb)
 {
 	struct scoutfs_sb_info *sbi = SCOUTFS_SB(sb);
 	struct scoutfs_lock *lock = sbi->rid_lock;
 	struct scoutfs_key key;
 	struct scoutfs_key last;
+	unsigned int deleted = 0;
+	u64 seq;
```
zabbo, line 1576: This is pulled from the item value and needs to be __le64, maybe also call it last_seq to match the orphan creation side?

kmod/src/inode.c
```diff
@@ -1595,10 +1602,12 @@ int scoutfs_scan_orphans(struct super_block *sb)
 			ret = -ENOENT;
 			break;
 		}
+
+		deleted++;
```
zabbo, line 1606: I know it sounds crazy, but I don't see anything that stops it from being possible to delete billions of files.  We're in that world.  Just make it a return 1 if it made progress, the caller doesn't care about the count at all.  And while we're at it maybe limit this to deleting a few thousand files per scheduled work?  It'll keep chain scheduling while there's items, but we want unmount to be able to get in without it having to chew through a potentially enormous amount of orphan items.

tests/funcs/fs.sh
```diff
@@ -17,6 +17,16 @@ t_sync_seq_index()
...
 }
 
+#
+# Ensure async inode deletion has run and completed.
+#
+t_sync_deleted()
+{
+	t_sync_seq_index
+	t_sync_seq_index
+	sleep 10
```
zabbo, line 27: The sleep feels both too long for the expected case, and maybe not long enough depending on what previous tests might have done.  If we're using this in a test that is specifically trying to kick the work, maybe we can add counters for the work .. finishing?  And use those to see when the work has executed since the unlink and so we can assume that the other counters reflect the unlink we're testing?

tests/tests/cross-mount-inode-del.sh
```diff
@@ -0,0 +1,51 @@
...
+
+sum_counters()
+{
+    local counter="$1"
+
+    cat /sys/fs/scoutfs/*/counters/$counter | awk '{sum += $1} END {print sum}'
+}
+
+echo "== open a file and delete it cross-mount"
+df_2_a=$(sum_counters forest_oino_deletable_false_2)
```
zabbo, line 15: If we're going to use counters to isolate our unlink, I think we have to stabilize for previous tests?  We could be able to look at the counters to see that the work execution counter changed without any per-orphan counters changing to see that the orphan work has drained?

zabbo (changes requested):
OK, this is what I've seen so far!

kmod/src/forest.c
```diff
@@ -782,7 +791,7 @@ int scoutfs_forest_oino_alloc(struct super_block *sb, u64 trans_seq)
 	bb->hdr.magic = cpu_to_le32(SCOUTFS_BLOCK_MAGIC_BLOOM);
 	bb->hdr.fsid = SCOUTFS_SB(sb)->super.hdr.fsid;
 	bb->hdr.blkno = cpu_to_le64(blkno);
-	bb->hdr.seq = cpu_to_le64(trans_seq);
+	bb->hdr.seq = cpu_to_le64(0);
```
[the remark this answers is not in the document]
agrover: We're setting hdr.seq to last_seq down in oino_update(), before put()ing the block. Since this value is overwritten before persistence maybe we shouldn't even be setting it here? if I'm understanding how this all works?

kmod/src/inode.c
agrover: Renaming scan_orphans to delete_orphans.

OK. I'm going to add a param to the fn for max orphans to delete, because I think we want that value coming from the oino code, not in this fn or inode.c in general. Inital max value 4096.

kmod/src/forest.c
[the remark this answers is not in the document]
agrover: ok so bloom.hdr.seq gets trans_seq and our new bloom.last_seq field gets last_seq.

kmod/src/forest.c
```diff
@@ -801,23 +810,162 @@ int scoutfs_forest_oino_alloc(struct super_block *sb, u64 trans_seq)
...
+			}
+			scoutfs_btree_put_iref(&iref);
+		}
+		if (ret < 0) {
+			if (ret == -ENOENT)
+				break;
+			goto out; /* including stale */
+		}
+
+		if (lt.oino_bloom_ref.blkno == 0)
```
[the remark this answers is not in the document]
agrover: That is a little annoying. I think I proposed something before and it was bad for a reason that I neglected to write down alas --> why not ask the server for # of clients or even a list of rids of all clients and use that to abort if # of blooms is too few/don't match?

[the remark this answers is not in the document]
zabbo: The number of clients isn't interesting, that doesn't tell you anything about the relative ages of the clients or the files they may have open or not. And the list of rids doesn't help either.   If you have all the rids, but one of them has blkno == 0, then we're back to where we started.

The problem is that blkno == 0 both when 1) you've first mounted and have opened things but haven't committed anything yet, and 2) you've been evicted and the oino bloom block has been freed.

In #1 we don't know *when* they mounted so we don't know if they were able to open files that have since been unlinked.  In #2 they're gone and we don't care and can ignore them.

I think the solution to all this is to create a valid empty oino bloom block with the client's first log_trees item that is created by the server when get_trees doesn't find one.  This is done early enough in mount that the client won't have any files open yet.  If we initialize last_seq in the initial oino bloom then we know that clients always have an oino bloom block in their log_trees item and we don't have to do any special nonsense when deciding if it's safe to delete.  Just iterate over the items and merge what we see.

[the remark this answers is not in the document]
agrover: OK I'll give that a go.

agrover:
changes:
* Initial oino bloom block now allocated by server.
* Refactor allocation since it's happening in two spots now

TODO:
* use existing t_counter_diff for cross-mount-inode-del test instead of home-grown
* look at reducing rather long sleep times in t_sync_deleted

agrover:
changes:
* add a forest/orphan_list_present_hint sysfs entry, that basically tells us if there are entries in the orphan list. Tests that rely on all orphans being deleted at a certain point check this. Also, inode deletion test uses this to verify an open but unlinked file remains on the orphan list until closed.

TODO:
* still seeing sporadic xfstest failures but :man_shrugging: ?

agrover:
Can I merge this?

agrover:
Rebased on top of latest `main`, with changes we made to use new block_dirty_ref() function.

agrover:
retest

kmod/src/inode.c
```diff
@@ -1541,8 +1542,6 @@ void scoutfs_evict_inode(struct inode *inode)
```
zabbo, line 1542: As we remove the inode destruction from evict_inode, we should remove the comment that talks about removing the persistent items.

kmod/src/inode.c
```diff
@@ -1559,33 +1558,42 @@ int scoutfs_drop_inode(struct inode *inode)
...
 
-	trace_scoutfs_scan_orphans(sb);
+	trace_scoutfs_delete_orphans(sb, bb, max);
 
 	init_orphan_key(&key, sbi->rid, 0);
 	init_orphan_key(&last, sbi->rid, ~0ULL);
 
-	while (1) {
-		ret = scoutfs_item_next(sb, &key, &last, NULL, 0, lock);
+	while (deleted < max) {
```
zabbo, line 1585: It looks like this'll always consume the full list because it's comparing the deleted bool to the max.  (And max is an unfortunate collision with the min/max macros.. maybe limit?).  I guess we want to max/limit to be applied to items scanned, not necessarily deleted, because we're worried about unmount hanging while this work is busy.  And undeletable inodes should be rare, so we're not too worried about constantly finding the entire front of the list undeletable.

kmod/src/inode.c
```diff
@@ -1595,10 +1603,15 @@ int scoutfs_scan_orphans(struct super_block *sb)
 			ret = -ENOENT;
 			break;
 		}
+
+		deleted = true;
```
zabbo, line 1607: I think we should only return 1 when we've made progress by delete_inode_items() returning success.  We probably don't want it to spin constantly if the list only contains failures?  It's a tricky case.

kmod/src/inode.c
```diff
@@ -1595,10 +1603,15 @@ int scoutfs_scan_orphans(struct super_block *sb)
 			ret = -ENOENT;
```
zabbo, line 1603: This existing test for wrapping of the ino in the key is kind of nonsense, but it's made even more nonsensical by adding the add_cpu/continue above without testing it.  I think it'd be cleaner to use a for(;;scoutfs_key_inc(&key)) instead of the manual key adds so we can remove both of them and this wrapping test.

kmod/src/inode.c
```diff
@@ -1609,13 +1622,18 @@ int scoutfs_orphan_inode(struct inode *inode)
...
 	trace_scoutfs_orphan_inode(sb, inode);
 
 	init_orphan_key(&key, sbi->rid, scoutfs_ino(inode));
 
-	ret = scoutfs_item_create(sb, &key, NULL, 0, lock);
+	trans_seq = cpu_to_le64(sbi->trans_seq);
+
+	ret = scoutfs_item_create(sb, &key, &trans_seq, sizeof(trans_seq), lock);
+
+	scoutfs_forest_oino_new_orphan(sb);
```
zabbo, line 1636: We probably only want to call new_orphans if the create succeeds.

kmod/src/server.c
```diff
@@ -380,6 +380,79 @@ static int alloc_move_empty(struct super_block *sb,
...
+	err = scoutfs_btree_next(sb, &super->trans_seqs, &key, &iref);
+	if (err == 0) {
+		key = *iref.key;
+		scoutfs_btree_put_iref(&iref);
+		tmp_last_seq = key.skts_trans_seq;
+
+	} else if (err == -ENOENT) {
+		tmp_last_seq = super->next_trans_seq;
+	} else {
+		return err;
```
zabbo, line 414: This is returning with the server->seq_rwsem held.  I'd change it to use ret instead of err, set ret = 0 in the -enoent case.  Have the seq_rwsem protect all that put the tmp_last_seq add -1 and setting the callers last_seq outside in an if (ret == 0) block and then return ret.

kmod/src/server.c
```diff
@@ -439,6 +512,8 @@ static int server_get_log_trees(struct super_block *sb,
 		memset(&lt, 0, sizeof(lt));
 		lt.rid = key.sklt_rid;
 		lt.nr = key.sklt_nr;
+
+		oino_newroot_alloc(sb, &lt.oino_bloom_ref);
```
zabbo, line 516: We should check for errors here.  It looks like if (ret < 0) goto unlock; should be fine.

kmod/src/server.c
```diff
@@ -627,6 +702,14 @@ static int reclaim_log_trees(struct super_block *sb, u64 rid)
 					&lt.meta_avail) ?:
 	      alloc_move_empty(sb, &super->data_alloc, &lt.data_avail) ?:
 	      alloc_move_empty(sb, &super->data_alloc, &lt.data_freed);
+
+	if (lt.oino_bloom_ref.blkno && !ret) {
+		ret = scoutfs_free_meta(sb, &server->alloc, &server->wri,
+					le64_to_cpu(lt.oino_bloom_ref.blkno));
+		BUG_ON(ret);
+		lt.oino_bloom_ref.blkno = 0;
```
zabbo, line 710: This should clear the whole ref struct, not just the blkno.

kmod/src/server.c
```diff
@@ -802,24 +870,7 @@ static int server_get_last_seq(struct super_block *sb,
...
-		scoutfs_btree_put_iref(&iref);
-		last_seq = key.skts_trans_seq;
-
-	} else if (ret == -ENOENT) {
-		last_seq = super->next_trans_seq;
-		ret = 0;
-	}
-
-	le64_add_cpu(&last_seq, -1ULL);
-	trace_scoutfs_trans_seq_last(sb, rid, le64_to_cpu(last_seq));
```
zabbo, line 820: It looks like the extraction of get_last_seq into its own function dropped this trace.

tests/tests/cross-mount-inode-del.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "orphan list present $(cat $(t_sysfs_path 1)/forest/orphan_list_present_hint)"
+
+# make sure we can still read through fd on mount 0
+# This actually doesn't seem to close fd.
+cat <&42
+# close it for real
+exec 42>&-
+
+# now deletable but we don't know it until the orphans are processed again
+echo "orphan list present $(cat $(t_sysfs_path 1)/forest/orphan_list_present_hint)"
```
zabbo, line 26: Just an observation: I don't think there's anything stopping the orphan list from being processed before we cat the hint after closing.  It'd be really hard to hit but we could get spurious test failures from it.  Something to try and remember and keep an eye out for, I guess.

kmod/src/forest.c
```diff
@@ -605,14 +662,282 @@ int scoutfs_forest_setup(struct super_block *sb)
...
+		return 0;
+
+	bb = finf->oino_bl->data;
+
+	ret = scoutfs_client_get_last_seq(sb, &last_seq);
+	if (ret)
+		return ret;
+
+	bb->last_seq = cpu_to_le64(last_seq);
+	finf->our_log.oino_bloom_ref.seq = bb->hdr.seq;
```
zabbo, line 796: This shouldn't be overwriting the ref seq.  _dirty_ref already set the ref for it back in newtrans_alloc.

kmod/src/forest.c
```diff
@@ -605,14 +662,282 @@ int scoutfs_forest_setup(struct super_block *sb)
...
+	DECLARE_FOREST_INFO(sb, finf);
+
+	finf->orphan_list_present_hint = true;
+}
+
+/*
+ * At the start of the transaction, alloc and mark dirty a meta block in
+ * anticipation of updating the OINO BF at the end of the transaction.
+ */
+int scoutfs_forest_oino_newtrans_alloc(struct super_block *sb, u64 trans_seq)
```
zabbo, line 748: This trans_seq argument isn't used anymore, we should drop it.

kmod/src/forest.c
```diff
@@ -605,14 +662,282 @@ int scoutfs_forest_setup(struct super_block *sb)
...
+
+		bl = read_bloom_ref(sb, &lt.oino_bloom_ref);
+		if (IS_ERR(bl)) {
+			ret = PTR_ERR(bl);
+			goto out;
+		}
+		bb = bl->data;
+
+		if (!tmp_bb) {
+			tmp_bb = vmalloc(SCOUTFS_BLOCK_LG_SIZE);
```
zabbo, line 863: We should just unconditionally allocate this up at the top of the function.  Then we don't need to test it when trying to delete, and it won't add a leak of the bl ref with the goto out without the put.  And you don't need to memset the block to 0, you can use the vmalloc that takes flags and ask it to zero.. something like __vmalloc(bytes, GFP_NOFS | __GFP_ZERO, PAGE_KERNEL);   (oh, which also reminds me that this probably wants GFP_NOFS so it can't recurse back into the fs :))

kmod/src/forest.c
```diff
@@ -605,14 +662,282 @@ int scoutfs_forest_setup(struct super_block *sb)
...
+			tmp_bb->last_seq = cpu_to_le64(U64_MAX);
+		}
+
+		/* Use oldest block's seq */
+		tmp_bb->last_seq = cpu_to_le64(
+			min(le64_to_cpu(bb->last_seq),
+			    le64_to_cpu(tmp_bb->last_seq)));
+
+		bitmap_or((long unsigned int *) &tmp_bb->bits,
+			  (const long unsigned int *) &tmp_bb->bits,
```
zabbo, line 878: unsigned long, not long unsigned int.

kmod/src/forest.c
```diff
@@ -605,14 +662,282 @@ int scoutfs_forest_setup(struct super_block *sb)
...
+			  (const long unsigned int *) &bb->bits,
+			  SCOUTFS_FOREST_BLOOM_BITS);
+
+		scoutfs_block_put(sb, bl);
+	}
+
+	if (tmp_bb) {
```
zabbo, line 885: (if we always vmalloced we could drop this test)

zabbo (changes requested):

kmod/src/inode.c
agrover: yeah this was bad logic left over from a previous version.

kmod/src/inode.c
agrover: nice!

tests/tests/cross-mount-inode-del.sh
agrover: I hear that.

kmod/src/server.c
agrover: nice catch!

kmod/src/server.c
agrover: ok.. I'm gonna just set blkno and seq to 0 instead of using memset. lmk if you'd rather memset or some other struct initialize-to-0 syntax.

kmod/src/forest.c
agrover: ok cool. also using GFP_ZERO on the vmalloced array of atomics, that also seems kosher.

kmod/src/server.c
agrover: :+1:

kmod/src/forest.c
agrover: ok, omitting that line.

agrover:
Rebased. All issues from review addressed.

versity-github:
Build 47 ${currentBuild.currentResult}

agrover:
obsolete

PR #18, opened 2021-02-15
Zab/quorum slots unmount

zabbo:


utils/man/scoutfs.8
```diff
@@ -57,18 +57,17 @@ a faster block device for the metadata device.
...
+Each \-Q option configures a quorum slot.  The NR specifies the number
+of the slot to configure, must be between 0 and 14, and only specified
+by one option.  ADDR and PORT are the numerical IPv4 address and port
+which will be used as the UDP endpoint for leader elections and as the
+TCP listening address for server connections.  The number of configured
+slots determines the size of the quorum of member mounts which must be
+present to start the server for the filesystem to operate.  A simple
+majority is typically required, while 1 mount is sufficient if only one
+or two slots are configured.  Until the majority quorum are present, all
+mounts will hang waiting for a server to connect to.
```
agrover, line 70: putting evil user hat on, can these be specified out of order? Can they be non consecutive? Repeating a NR is an error, I think I saw somewhere, yes?

kmod/src/quorum.c
```diff
@@ -13,766 +13,1073 @@
...
-		memset(our_blk, 0, SCOUTFS_BLOCK_SM_SIZE);
+	/* start out as a follower */
+	qst.role = FOLLOWER;
+	qst.term = le64_to_cpu(blk.term);
+	qst.vote_for = -1;
+	qst.vote_bits = 0;
 
-		scoutfs_inc_counter(sb, quorum_cycle);
+	/* see if there's a server to chose heartbeat or election timeout */
+	if (scoutfs_quorum_server_sin(sb,&unused) == 0)
```
agrover, line 557: typo, add space

kmod/src/quorum.c
```diff
@@ -13,766 +13,1073 @@
...
+		if (!test_bit(QINF_FLAG_SERVER, &qinf->flags) &&
+		    qst.role == LEADER) {
+			qst.role = FOLLOWER;
+			qst.vote_for = -1;
+			qst.vote_bits = 0;
+			qst.timeout = election_timeout();
+			scoutfs_inc_counter(sb, quorum_server_shutdown);
+
+			send_msg_others(sb, SCOUTFS_QUORUM_MSG_RESIGNATION,
+					qst.term);
```
agrover, line 590: Do we need resignation?

agrover (changes requested):
initial pass with questions and small stuff.

utils/src/parse.c
```diff
@@ -152,3 +155,56 @@ int parse_timespec(char *str, struct timespec *ts)
...
+{
+#define ADDR_CHARS 45 /* max ipv6 */
+	char addr[ADDR_CHARS + 1] = {'\0',};
+	struct in_addr in;
+	int port;
+	int parsed;
+	int nr;
+	int ret;
+
+	ret = sscanf(arg, "%u:%"__stringify(ADDR_CHARS)"[^:]:%u%n",
```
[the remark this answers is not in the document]
zabbo: Given that I had to squint at the man page to get it to work, I definitely agree that I should have left a helpful comment :).

Hmm, yeah, that's a good point, ipv6 parsing is already going to be a nightmare and using : as our separator is not doing us any favors.  I'll fix it up to use ",", that's a good suggestion.  I think we're close, yeah, but I'm guessing that testing will be pretty obnoxious.

kmod/src/server.c
```diff
@@ -1030,15 +1030,20 @@ static int insert_mounted_client(struct super_block *sb, u64 rid,
...
 
 	init_mounted_client_key(&key, rid);
 	mcv.flags = 0;
 	if (gr_flags & SCOUTFS_NET_GREETING_FLAG_QUORUM)
 		mcv.flags |= SCOUTFS_MOUNTED_CLIENT_QUORUM;
 
-	return scoutfs_btree_insert(sb, &server->alloc, &server->wri,
-				    &super->mounted_clients, &key, &mcv,
-				    sizeof(mcv));
+	mutex_lock(&server->farewell_mutex);
```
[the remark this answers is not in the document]
zabbo: Heh, yeah, I thought the same thing but didn't want to add churn to the patch.  I'll layer on a quick rename patch, easy peasy.

README.md
```diff
@@ -65,8 +65,13 @@ The steps for getting scoutfs mounted and operational are:
...
 
-In this example we run all of these commands on three nodes.  The names
-of the block devices are the same on all the nodes.
+In this example we use three nodes.  The names of the block devices are
+the same on all the nodes.  Two of the nodes will be quorum members.  A
+majority of quorum members must be mounted to elect a leader to run a
+server that all the mounts connect to.  It should be noted that two
+quorum members results in a majority of one, each member itself, so
+split brain elections are possible but so unlikely that it's fine for a
+demonstration.
```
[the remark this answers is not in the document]
zabbo: It's perfectly reasonable to want to run with just two mounts, say for development testing or perf exploration or whatever.  I'm fine with this little mention in the readme, maybe we could use a more thorough exploration of the risk in some documentation of the quorum/clustering bits somewhere.

utils/man/scoutfs.8
zabbo: yes, yes, correct :).  I'll try to say so specifically in the man page.

kmod/src/quorum.c
zabbo: I think so.  Without it quorum members can't differentiate between a server dying, so they wait for the longer heartbeat timeout, and a server exiting cleanly, at which point they should use the shorter election timeout.  Without it umount of the leader would cause a heartbeat timeout delay in service until the next leader is elected.  The mount/unmount test would barely do anything, it would be spending all its time sitting around waiting.  This is one of the effects of our different model than the share-nothing raft key/value store.. we're using the leader election to stand up and shut down an exclusive service.

README.md
```diff
@@ -88,24 +93,29 @@ of the block devices are the same on all the nodes.
...
    ```
 
 3. Mount the Filesystem
 
-   Each mounting node provides its local IP address on which it will run
-   an internal server for the other mounts if it is elected the leader by
-   the quorum.
+   First, mount each of the quorum nodes so that they can elect and
+   start a server for the remaining node to connect to.  The slot
+   numbers were specified with "X:" in the -Q options to mkfs.
```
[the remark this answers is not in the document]
zabbo: Agreed, I made it a bit clearer.

utils/src/parse.c
```diff
@@ -152,3 +155,65 @@ int parse_timespec(char *str, struct timespec *ts)
...
+{
+#define ADDR_CHARS 45 /* max ipv6 */
+	char addr[ADDR_CHARS + 1] = {'\0',};
+	struct in_addr in;
+	int port;
+	int parsed;
+	int nr;
+	int ret;
+
+	/* leading and trailing ints, an inner sized string without :, all separated by : */
```
agrover, line 177: separated by "," now, right?

utils/man/scoutfs.8
```diff
@@ -57,18 +57,20 @@ a faster block device for the metadata device.
...
+of the slot to configure which must be between 0 and 14.  Each slot
+number must only be used once, but they can be used in any order and
+they need not be consecutive.  This is to allow natural relationships
+between slot numbers and nodes which may have arbitrary numbering
+schemes.  ADDR and PORT are the numerical IPv4 address and port which
+will be used as the UDP endpoint for leader elections and as the TCP
+listening address for server connections.  The number of configured
+slots determines the size of the quorum of member mounts which must be
+present to start the server for the filesystem to operate.  A simple
+majority is typically required, while 1 mount is sufficient if only one
```
agrover, line 71: "one" instead of "1" here

agrover (commented):
2 last little things but please go ahead and merge whenever. Looks good.

zabbo:
> 2 last little things but please go ahead and merge whenever. Looks good.

Cool, I fixed those up and am pressing the buttons!

PR #20, opened 2021-03-02
Zab/remove trans spinlock

zabbo:


kmod/src/alloc.h
```diff
@@ -72,7 +72,10 @@
  * transaction.
  */
 struct scoutfs_alloc {
+	/* protects modifying avail/free list_heads during transactions */
 	spinlock_t lock;
+	/* write nests in spinlock, allows read-mostly reading of avail/free list_heads */
+	seqcount_t seqcount;
```
agrover, line 78: This useful pattern looks like an open-coded version of [seqlock](https://elixir.bootlin.com/linux/v3.10.108/source/include/linux/seqlock.h#L174)?

zabbo: Huh, yeah, look at that!  I didn't realize there was an implementation of this.  I'll rework it a bit to use that :).

agrover:
that's all I got. Merge when ready.

PR #22, opened 2021-03-06
Reserve space in superblock for IPv6 addresses

agrover:
Reserve 16 bytes by making scoutfs_inet_addr an array[4] of __le32.

Also add a tag field to differentiate between v4 and v6 usage.

Change code since addr is now an array.

### DIFF
--- a/kmod/src/format.h
+++ b/kmod/src/format.h
@@ -86,11 +86,33 @@ struct scoutfs_timespec {
 	__u8 __pad[4];
 };
 
-/* XXX ipv6 */
-struct scoutfs_inet_addr {
-	__le32 addr;
+enum scoutfs_inet_family {
+	SCOUTFS_AF_NONE = 0,
+	SCOUTFS_AF_IPV4 = 1,
+	SCOUTFS_AF_IPV6 = 2,
+};
+
+struct scoutfs_inet_addr4 {
+	__le16 family;
 	__le16 port;
-	__u8 __pad[2];
+	__le32 addr;
+};
+
+/*
+ * Not yet supported by code.
+ */
+struct scoutfs_inet_addr6 {
+	__le16 family;
+	__le16 port;
+	__u8 addr[16];
+	__le32 flow_info;
+	__le32 scope_id;
+	__u8 __pad[4];
+};
+
+union scoutfs_inet_addr {
+	struct scoutfs_inet_addr4 v4;
+	struct scoutfs_inet_addr6 v6;
 };
 
 /*
@@ -591,7 +613,7 @@ struct scoutfs_quorum_message {
 struct scoutfs_quorum_config {
 	__le64 version;
 	struct scoutfs_quorum_slot {
-		struct scoutfs_inet_addr addr;
+		union scoutfs_inet_addr addr;
 	} slots[SCOUTFS_QUORUM_MAX_SLOTS];
 };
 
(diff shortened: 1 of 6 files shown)


kmod/src/format.h
```diff
@@ -86,11 +86,15 @@ struct scoutfs_timespec {
...
+ * ipv6 support (TBD) will use all 4 and set is_ipv6_addr.
+ */
 struct scoutfs_inet_addr {
-	__le32 addr;
+	__le32 addr[4];
 	__le16 port;
-	__u8 __pad[2];
+	__u8 is_ipv6_addr;
+	__u8 __pad[5];
 };
```
zabbo, line 98: By the principle of least surprise, let's make this look like the sockaddr union in the kernel.  It'd be 16 bytes instead of 24 and the family would differentiate between whatever types instead of having a u8 boolean.  I suspect we'll want to have the current code use the family to recognize ipv4 to make the eventual addition of ipv6 a lot cleaner, but it's a bit hazy.  Maybe we should have a chat about it.

zabbo (changes requested):

agrover:
Union on the outside!

kmod/src/net.h
```diff
@@ -90,19 +90,20 @@ enum conn_flags {
...
 
-static inline void scoutfs_addr_from_sin(struct scoutfs_inet_addr *addr,
+static inline void scoutfs_addr_from_sin(union scoutfs_inet_addr *addr,
 					 struct sockaddr_in *sin)
 {
-	addr->addr = be32_to_le32(sin->sin_addr.s_addr);
-	addr->port = be16_to_le16(sin->sin_port);
-	memset(addr->__pad, 0, sizeof(addr->__pad));
+	memset(addr, 0, sizeof(union scoutfs_inet_addr));
+	addr->v4.family = sin->sin_family;
```
zabbo, line 104: We can't rely on the family constants being consistent across ABIs.  We need our own family enum and a table to go between the two, sort of like we do for transporting errnos as message error responses.  But hilariously enough, this function isn't used anymore!  It was only used to store the server's bound address in the super for clients to connect to.  Now the clients connect to known addresses written by mkfs.  The kernel never writes addresses into the quorum config.  So we can just remove this function.

utils/src/mkfs.c
```diff
@@ -395,22 +395,22 @@ static bool valid_quorum_slots(struct scoutfs_quorum_slot *slots)
 	int j;
 
 	for (i = 0; i < SCOUTFS_QUORUM_MAX_SLOTS; i++) {
-		if (slots[i].addr.addr == 0)
+		if (slots[i].addr.v4.addr == 0)
 			continue;
```
zabbo, line 400: The only persistent address inputs are the config slots.  This should verify that the family is known (only v4 today) and fail if it sees an unrecognized family.  This will ensure that the structures are already consistent when they're found for the first time by code that supports both v4 and v6.

utils/src/parse.c
```diff
@@ -213,7 +213,7 @@ int parse_quorum_slot(struct scoutfs_quorum_slot *slot, char *arg)
 		return -EINVAL;
 	}
```
zabbo, line 215: This needs to set the family to our v4 enum.

zabbo (changes requested):
Just a few tweaks to initialize and test the new family field.  While we're defining the enum let's make an explicit _NULL=0 value so that we can see addrs that are specifically initialized by empty.  (So some day we can refactor all those .addr == 0 tests to be some .family == NULL helper, but that doesn't need to happen here.)

agrover:
retest

versity-github:
Build 46 ${BUILD_STATUS}

PR #28, opened 2021-03-31
scoutfs: introduce CodingStyle.txt

agrover:
Add a coding style document that tries to record the conventions that
the project uses.  It seemed more appropriate to put it up in the -kmod
git repo context rather than in src/ which would end up in fs/scoutfs/
upstream.

### DIFF
--- a/CodingStyle.txt
+++ b/CodingStyle.txt
@@ -1,0 +1,82 @@
+
+We try to maintain a consistent coding style across the project.  It's
+admitedly arbitrary and starts with and is based on upstream's
+Documentation/CodingStyle.  Conventions are added here as they come up
+during review.  We'll demonstrate each sylistic preference with a diff
+snippet.
+
+== Try to make one exit point for reasonably long functions
+
+{
+-	void *a;
+-	void *b;
++	void *a = NULL;
++	void *b = NULL;
++	int ret;
+
+	a = kalloc();
+-	if (!a)
+-		return 1;
++	if (!a) {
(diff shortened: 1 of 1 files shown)


agrover:
[copying PR from PR against old repo](https://github.com/versity/scoutfs-kmod-dev/pull/39) I think the only question is exactly where to put it. Why not toplevel?

versity-github:
Build 52 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/52/ git: 8ee41caa24f89e748e17836103eb7ea356b523a4

zabbo:
> [copying PR from PR against old repo](https://github.com/versity/scoutfs-kmod-dev/pull/39) I think the only question is exactly where to put it. Why not toplevel?

Well, I'd first ask if we want it to begin with.  It landed with a resounding thud when I first threw it together.  Is it needed?

agrover:
Yeah I think it's a good idea. To just have a place to put style preferences. We may come up with more.

 I think I just read the PR and got the info that way but that's kinda cheating or something.

versity-github:
Build 6 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/6/ git: 8ee41caa24f89e748e17836103eb7ea356b523a4

versity-github:
Can one of the admins verify this patch?

versity-github:

zabbo:
This is ancient and I don't think we'll be doing this.

PR #29, opened 2021-03-31
Cleanup

agrover:
Some old trivial cleanup commits I had lying around. I think they look right?

### DIFF
--- a/kmod/src/forest.c
+++ b/kmod/src/forest.c
@@ -276,7 +276,6 @@ int scoutfs_forest_read_items(struct super_block *sb,
 	scoutfs_inc_counter(sb, forest_read_items);
 	calc_bloom_nrs(&bloom, &lock->start);
 
-	roots = lock->roots;
 retry:
 	ret = scoutfs_client_get_roots(sb, &roots);
 	if (ret)
--- a/kmod/src/forest.c
+++ b/kmod/src/forest.c
@@ -348,15 +348,9 @@ retry:
 	ret = 0;
 out:
 	if (ret == -ESTALE) {
-		if (memcmp(&prev_refs, &refs, sizeof(refs)) == 0) {
-			ret = -EIO;
-			goto out;
-		}
+		if (memcmp(&prev_refs, &refs, sizeof(refs)) == 0)
+			return -EIO;
 		prev_refs = refs;
-
-		ret = scoutfs_client_get_roots(sb, &roots);
-		if (ret)
-			goto out;
 		goto retry;
 	}
 
--- a/kmod/src/format.h
+++ b/kmod/src/format.h
@@ -259,7 +259,7 @@ struct scoutfs_btree_block {
 	__le16 mid_free_len;
 	__u8 level;
 	__u8 __pad[7];
-	struct scoutfs_btree_item items[0];
+	struct scoutfs_btree_item items[];
 	/* leaf blocks have a fixed size item offset hash table at the end */
 };
 
@@ -307,7 +307,7 @@ struct scoutfs_alloc_list_block {
 	struct scoutfs_block_ref next;
 	__le32 start;
 	__le32 nr;
-	__le64 blknos[0]; /* naturally aligned for sorting */
+	__le64 blknos[]; /* naturally aligned for sorting */
 };
 
 #define SCOUTFS_ALLOC_LIST_MAX_BLOCKS					      \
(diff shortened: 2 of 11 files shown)


versity-github:
Build 53 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/53/ git: 4c0cf47c0accd78e82a9f5b274ee895fd133b689

kmod/src/forest.c
```diff
@@ -348,15 +348,9 @@ int scoutfs_forest_read_items(struct super_block *sb,
 	ret = 0;
 out:
 	if (ret == -ESTALE) {
-		if (memcmp(&prev_refs, &refs, sizeof(refs)) == 0) {
-			ret = -EIO;
-			goto out;
-		}
+		if (memcmp(&prev_refs, &refs, sizeof(refs)) == 0)
+			return -EIO;
```
zabbo, line 352: Yeah, making it match the other retry handling makes sense, though I don't really like constructions with multiple returns when we can avoid it.

kmod/src/lock.c
```diff
@@ -667,7 +667,6 @@ static void lock_grant_worker(struct work_struct *work)
 		lock->request_pending = 0;
 		lock->mode = nl->new_mode;
 		lock->write_version = le64_to_cpu(nl->write_version);
-		lock->roots = gr->roots;
```
zabbo, line 670: Keep going, its a chain of single uses being removed:
 - go back to grant_reply using net_lock struct, remove grant_reply struct
 - lock_server doesn't need to fill grant roots, so scoutfs_server_get_fs_roots can go back to just being a private get_fs_roots (to match set_fs_roots) which is only called by the message handler.

You can see this stuff being introduced in 
```
commit b7943c5412ba7995f6205fac07cedf36a1fc652f
Author: Zach Brown <zab@versity.com>
Date:   Thu May 7 11:21:27 2020 -0700

    scoutfs: avoid reading forest roots with block IO

```

zabbo (changes requested):
Nice, yeah, the list_head and retry reworks are fine.

The removal of the lock roots is probably good on balance just to simplify a bit.   It was an optimization to avoid the message round trip as long as you don't see retries, but it was lost as the forest was reworked for the item cache.  But the lock roots can be very stale indeed and we were never updating the roots stored in the lock as we got more recent versions.  It's probably eaiser to always try recent roots, and if we wanted to add more optimizations here in the future we could -- and cache the newly acquired root in the lock.

In any case, there's a bit more to dangling unused stuff to remove if we're tearing out the lock roots.

versity-github:
Build 58 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/58/ git: 7dcb5b34b6f799860ec352ca39b71dd2fbafbb54

versity-github:
Build 59 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/59/ git: 91913558e44cdebef0dee127f8d03dbe28902de0

versity-github:
Build 60 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/60/ git: 0102b4314ee71e740ea47b5667f3d7a0e3c0ef26

kmod/src/format.h
```diff
@@ -897,7 +897,6 @@ struct scoutfs_net_lock {
 
 struct scoutfs_net_lock_grant_response {
```
zabbo, line 898: Now that the roots are removed this _grant_response struct is a meaningless wrapper around _net_lock.  Everywhere that the _grant_response struct is used can be replaced with _net_lock.  You can see this in the function prototypes in client.h and server.h for {client,server}_{request,response}.   The grant was an outlier only because it included the roots.  With those gone, it can go back to matching the other three core lock messages that cross the wire by sending the _net_lock struct.  That's what I was trying to describe with the first bullet in the initial "keep going" review comment.

zabbo (changes requested):
Nice, I learned about the modern process/deprecated doc today!  (Goodness, I need to re-engage with upstream.)

kmod/src/format.h
agrover: cool, will do.

versity-github:
Build 61 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/61/ git: 4495bf1a87702782b72b171523734af5f3cc913a

versity-github:
Build 62 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/62/ git: 4630b77b45751fe5199214dfeccb420b84f1a64c

zabbo (approved):

zabbo:
Nice, thanks for cleaning up the commits!

PR #31, opened 2021-04-08
Block cache shrink restart waits for rcu callbacks

zabbo:
We're seeing cpu livelocks in block shrinking where counters show that a
single block cache shrink call is only getting EAGAIN from repeated
rhashtable walk attempts.  It occurred to me that the running task might
be preventing an RCU grace period from ending by never blocking.

The hope of this commit is that by waiting for rcu callbacks to run
we'll ensure that any pending rebalance callback runs before we retry
the rhashtable walk again.  I haven't been able to reproduce this easily
so this is a stab in the dark.

### DIFF
--- a/kmod/src/block.c
+++ b/kmod/src/block.c
@@ -1074,10 +1074,11 @@ restart:
 		if (bp == NULL)
 			break;
 		if (bp == ERR_PTR(-EAGAIN)) {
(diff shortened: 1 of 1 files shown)


zabbo:
This seems to have fixed the livelocks we were seeing in restore tests, so let's get this in.

PR #35, opened 2021-04-23
Handle back to back invalidation requests

zabbo:
The client's incoming lock invalidation request handler triggers a
BUG_ON if it gets a request for a lock that is already processing a
previous invalidation request.  The server is supposed to only send
one request at a time.

The problem is that the batched invalidation request handling will send
responses outside of spinlock coverage before reacquirin the lock and
finishing processing once the response send has been successful.

This gives a window for another invalidation request to arrive after the
response was sent but before the invalidation finished processing.  This
triggers the bug.

The fix is to mark the lock such that we can recognize a valid second
request arriving after we send the response but before we finish
processing.  If it arrives we'll continue invalidation processing with
the arguments from the new request.

### DIFF
--- a/kmod/src/lock.c
+++ b/kmod/src/lock.c
@@ -713,7 +713,9 @@ int scoutfs_lock_grant_response(struct super_block *sb,
 /*
  * Each lock has received a lock invalidation request from the server
  * which specifies a new mode for the lock.  The server will only send
- * one invalidation request at a time for each lock.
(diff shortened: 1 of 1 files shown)


versity-github:
Build 74 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/74/ git: a27c54568c325a30189b17755b2c751f7304e4c1

PR #36, opened 2021-04-23
Fix accidental EINVAL in move_blocks

zabbo:
When move blocks is staging it requires an overlapping offline extent to
cover the entire region to move.

It performs the stage by modifying extents at a time.  If there are
fragmented source extents it will modify each of them at a time in the
region.

When looking for the extent to match the source extent it looked from
the iblock of the start of the whole operation, not the start of the
source extent it's matching.  This meant that it would find a the first
previous online extent it just modified, which wouldn't be online, and
would return -EINVAL.

The fix is to have it search from the logical start of the extent it's
trying to match, not the start of the region.

### DIFF
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -1285,7 +1285,7 @@ int scoutfs_data_move_blocks(struct inode *from, u64 from_off,
 
 			if (is_stage) {
 				ret = scoutfs_ext_next(sb, &data_ext_ops, &to_args,
-						       to_iblock, 1, &off_ext);
+						       to_start, 1, &off_ext);
 				if (ret)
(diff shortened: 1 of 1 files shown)


versity-github:
Build 73 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/73/ git: 63148d426e32946e1e2fc415b200047690a45761

agrover:
retest

versity-github:
Build 75 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/75/ git: 63148d426e32946e1e2fc415b200047690a45761

PR #38, opened 2021-04-30
Zab/read xattr deadlocks

zabbo:


kmod/src/file.c
```diff
@@ -43,30 +48,32 @@ ssize_t scoutfs_file_aio_read(struct kiocb *iocb, const struct iovec *iov,
 	int ret;
 
 retry:
+	/* protect checked extents from release */
+	mutex_lock(&inode->i_mutex);
+	atomic_inc(&inode->i_dio_count);
+	mutex_unlock(&inode->i_mutex);
```
agrover, line 54: I know this code block is just moved from below, but... why do we need to lock a mutex to increment an atomic?

(I see the comment above fs/inode.c inode_dio_wait(), but I'm not sure it applies here or not.)

agrover (approved):
I had one Q but it's tangential, so this can be merged at will.

versity-github:
Build 79 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/79/ git: a5ca5ee36da6234e43baed93821e69cd6f3f3314

kmod/src/file.c
zabbo: dio_count waiters wait for the count to drop to zero before they continue on to do their work while holding the mutex.  It's built for truncate to wait for IO to drain before modifying extents (which could have IO in flight targeting those extents.).  If we just set the count a truncate could be doing work while still holding the mutex.  Setting the count under the mutex means that we ensure that no truncate can do work until we drop the count.  The next truncate to get the mutex will have to wait for us to drop the count.  (For us, release is just a truncate that leaves behind offline extents.)

PR #48, opened 2021-08-06
Queue invalidation during previous request

zabbo:
The locking protocol only allows one outstanding invalidation request
for a lock at a time.  The client invalidation state is a bit hairy and
involves removing the lock from the invalidation list while it is being
processed which includes sending the response.  This means that another
request can arrive while the lock is not on the invalidation list.  We
have fields in the lock to record another incoming request which puts
the lock back on the list.

But the invalidation work wasn't always queued again in this case.  It
*looks* like the incoming request path would queue the work, but by
definition the lock isn't on the invalidation list during this race.  If
it's the only lock in play then the invalidation list will be empty and
the work won't be queued.  The lock can get stuck with a pending
invalidation if nothing else kicks the invaliation worker.  We saw this
in testing when the root inode lock group missed the wakeup.

The fix is to have the work requeue itself after putting the lock back
on the invalidation list when it notices that another request came in.

### DIFF
--- a/kmod/src/lock.c
+++ b/kmod/src/lock.c
@@ -862,8 +862,9 @@ static void lock_invalidate_worker(struct work_struct *work)
 			lock->invalidate_pending = 0;
 			wake_up(&lock->waitq);
 		} else {
-			/* another request filled nl/net_id, put it back on the list */
+			/* another request filled nl/net_id, back on the list and requeue */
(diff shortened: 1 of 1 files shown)


versity-github:
Build 98 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/98/ git: 65ac42831f025d10d8d519d389475b4a99bfb081

PR #50, opened 2021-08-31
Verify dentries after locking

zabbo:
Our dir methods were trusting dentry args.  The vfs code paths use
i_mutex to protect dentries across revalidate or lookup and method
calls.  But that doesn't protect methods running in other mounts.
Multiple nodes can interleave the initial lookup or revalidate then
actual method call.

Rename got this right.  It is very paranoid about verifying inputs after
acquiring all the locks it needs.

We extend this pattern to the rest of the methods that need to use the
mapping of name to inode (and our hash and pos) in dentries.  Once we
acquire the parent dir lock we verify that the dentry is still current,
returning -EEXIST or -ENOENT as appropriate.

Along these lines, we tighten up dentry info correctness a bit by
updating our dentry info (recording lock coverage and hash/pos) for
negative dentries produced by lookup or as the result of unlink.

### DIFF
--- a/kmod/src/dir.c
+++ b/kmod/src/dir.c
@@ -316,6 +316,52 @@ out:
 	return ret;
 }
 
+/*
+ * Verify that the caller's dentry still precisely matches our dirent
+ * items.
+ *
+ * The caller has a dentry that the vfs revalidated before they acquired
+ * their locks.  If the dentry is still covered by a lock we immediately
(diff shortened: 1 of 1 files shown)


versity-github:
Build 100 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/100/ git: 79fbaa64815bd6cfdae0613dfcfe845751500599

PR #53, opened 2021-10-13
Fix mkdir-rename-rmdir test script

bgly:
The current script gets stuck in an infinite loop when the test
suite is started with 1 mount point. This is due to the advancement
part of the script in which is advances the ops for each mount.
There needs to be a check that ensures the NR_MNTS > 1 before it
starts to initiate the advancement of ops for each mount otherwise
break out of the original while loop since the ops are already
finished for the original mount.

Signed-off-by: Bryant Duffy-Ly <bduffyly@versity.com>


tests/tests/mkdir-rename-rmdir.sh
```diff
@@ -43,17 +43,22 @@ while [ ${op_mnt[$NR_OPS]} == 0 ]; do
 		fi
 	done
 
-	# advance through mnt nrs for each op
-	i=0
-	while [ ${op_mnt[$NR_OPS]} == 0 ]; do
```
zabbo, line 48: We don't need to special case the whole loop, just change the while condition to "$i -lt $NR_OPS", and it's done.  It's advancing through the combinations of mounts for each op until the final op wraps back around to 0 (all mounts must have also wrapped to 0), and it's back to the initial all 0s and can stop.

zabbo: Sure, bleh.   I guess I'd move the test from before the increments to after. 

```
diff --git a/tests/tests/mkdir-rename-rmdir.sh b/tests/tests/mkdir-rename-rmdir.sh
index 0379cae..529b9cb 100755
--- a/tests/tests/mkdir-rename-rmdir.sh
+++ b/tests/tests/mkdir-rename-rmdir.sh
@@ -23,9 +23,7 @@ else
        NR_MNTS=$T_NR_MOUNTS
 fi
 
-# test until final op mount dir wraps
-while [ ${op_mnt[$NR_OPS]} == 0 ]; do
-
+while : ; do
        # sequentially perform each op from its mount dir
        for op in $(seq 0 $((NR_OPS - 1))); do
                m=${op_mnt[$op]}
@@ -45,7 +43,7 @@ while [ ${op_mnt[$NR_OPS]} == 0 ]; do
 
        # advance through mnt nrs for each op
        i=0
-       while [ ${op_mnt[$NR_OPS]} == 0 ]; do
+       while [ $i -lt $NR_OPS ]; do
                ((op_mnt[$i]++))
                if [ ${op_mnt[$i]} -ge $NR_MNTS ]; then
                        op_mnt[$i]=0
@@ -54,6 +52,9 @@ while [ ${op_mnt[$NR_OPS]} == 0 ]; do
                        break
                fi
        done
+
+       # done when the last op's mnt nr wrapped
+       [ $i -ge $NR_OPS ] && break
 done
 
 t_pass
```

versity-github:
Build 104 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/104/ git: 2d1b8b0afa55dd302c6959aca3240c044399bb35

PR #54, opened 2021-10-21
Fix client/server abort conn on force unmount

bgly:
Currently when running certain test cases the code will
hang indefinitely due to forced unmount not clearing up
scoutfs_net_sync_request. The fix here is to call
scoutfs_net_shutdown early for the client and server in order
to clear up the pending message queues.

### DIFF
--- a/kmod/src/btree.c
+++ b/kmod/src/btree.c
@@ -1233,10 +1233,6 @@ static int btree_walk(struct super_block *sb,
 	    WARN_ON_ONCE((flags & (BTW_GET_PAR|BTW_SET_PAR)) && !par_root))
 		return -EINVAL;
 
-	/* all ops come through walk and walk calls all reads */
-	if (scoutfs_forcing_unmount(sb))
-		return -EIO;
-
 	scoutfs_inc_counter(sb, btree_walk);
 
 restart:
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -2426,7 +2426,9 @@ static int server_commit_log_merge(struct super_block *sb,
 	struct scoutfs_log_merge_range rng;
 	struct scoutfs_key key;
 	char *err_str = NULL;
-	int ret;
+	bool deleted = false;
+	int ret = 0;
+	int err = 0;
 
 	scoutfs_key_set_zeros(&rng.end);
 
@@ -2474,6 +2476,7 @@ static int server_commit_log_merge(struct super_block *sb,
 		err_str = "deleting orig request";
 		goto out;
 	}
+	deleted = true;
 
 	if (le64_to_cpu(comp->flags) & SCOUTFS_LOG_MERGE_COMP_ERROR) {
 		/* restore the range and reclaim the allocator if it failed */
@@ -2533,8 +2536,11 @@ out:
 	if (ret < 0)
 		scoutfs_err(sb, "error %d committing log merge: %s", ret, err_str);
 
-	ret = scoutfs_server_apply_commit(sb, ret);
-	BUG_ON(ret < 0); /* inconsistent */
+	err = scoutfs_server_apply_commit(sb, ret);
+	BUG_ON(ret < 0 && deleted); /* inconsistent */
+
+	if (ret == 0)
+		ret = err;
 
 	return scoutfs_net_response(sb, conn, cmd, id, ret, NULL, 0);
 }
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -2068,6 +2068,19 @@ static void server_log_merge_free_work(struct work_struct *work)
 			break;
 		}
 
+		/* Dirty the btree before freeing so that we can pin it
+		 * so that later touches will succeed.
+		 */
+		init_log_merge_key(&key, SCOUTFS_LOG_MERGE_FREEING_ZONE,
+				   le64_to_cpu(fr.seq), 0);
+		ret = scoutfs_btree_dirty(sb, &server->alloc,
+						&server->wri, &super->log_merge,
+						&key);
+		if (ret < 0) {
+			err_str = "dirtying log btree";
+			break;
+		}
+
 		ret = scoutfs_btree_free_blocks(sb, &server->alloc,
 						&server->wri, &fr.key,
 						&fr.root, 10);
@@ -2077,8 +2090,6 @@ static void server_log_merge_free_work(struct work_struct *work)
 		}
 
 		/* freed blocks are in allocator, we *have* to update key */
-		init_log_merge_key(&key, SCOUTFS_LOG_MERGE_FREEING_ZONE,
-				   le64_to_cpu(fr.seq), 0);
 		if (scoutfs_key_is_ones(&fr.key))
 			ret = scoutfs_btree_delete(sb, &server->alloc,
 						   &server->wri,
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -3823,6 +3823,7 @@ static void scoutfs_server_worker(struct work_struct *work)
 	struct scoutfs_net_connection *conn = NULL;
 	DECLARE_WAIT_QUEUE_HEAD(waitq);
 	struct sockaddr_in sin;
+	bool alloc_init = false;
 	u64 max_seq;
 	int ret;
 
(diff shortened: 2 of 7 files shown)


kmod/src/server.c
```diff
@@ -2124,7 +2124,7 @@ static void server_log_merge_free_work(struct work_struct *work)
 	int ret = 0;
 
 	/* shutdown waits for us, we'll eventually load set shutting_down */
-	while (!server->shutting_down) {
+	while (!server->shutting_down && !scoutfs_forcing_unmount(sb)) {
```
zabbo, line 2127: Where did this come from?  Fixing a hang?   Avoiding generating errors?

kmod/src/server.c
```diff
@@ -3989,6 +3990,15 @@ void scoutfs_server_abort(struct super_block *sb)
 	stop_server(server);
 }
 
+void scoutfs_server_abort_conn(struct super_block *sb)
+{
+	struct server_info *server = SCOUTFS_SB(sb)->server_info;
+
+	if (server->conn) {
```
zabbo, line 3997: It turns out this is dangerous because the server conn lifetime is different than the client.  For the client, we have a conn for the duration of the mount so it's a lot easier to reference the conn from within the mount.   For the server, the conn is allcoated and freed each time the server starts and stops.   It'd be pretty easy for this to deref a null conn.   It turns out all we should do is just call scoutfs_server_abort() from umount_begin, we don't need this _server_abort_conn at all, I don't think.

zabbo (changes requested):

kmod/src/server.c
bgly: Okay this makes sense now since I was trying to avoid the deref of a null in the above with:
`while (!server->shutting_down && !scoutfs_forcing_unmount(sb)) {`

versity-github:
Build 107 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/107/ git: 2354d04089de99062626e03ec7a8c1b7bef5aed4

versity-github:
Build 114 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/114/ git: 2240a61d4812ad3b424600ed0be5208c979b6875

versity-github:
Build 115 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/115/ git: eae763ca9c65634154b697f53b612afe6cd78898

zabbo (changes requested):
OK, I think this can be broken up into a few themes that can either be removed or separated into their own commits with summaries.

All the short circuiting with _forcing_unmount() calls can go.

Almost all of the btree_dirty calls can go.   There's the one attempt to avoid inconsistency in server_log_merge_free_work().   That could be its own commit if it was fixed up.

There's having the server call prepare_commit and forget its dirty blocks as it shuts down.   That'll be good once the extra forget calls are removed.

There's breaking stuck client sends out of waiting for network replies.   That'll be good with the loops over the send queues removed and replaced with a call into the client to call net_shutdown() if it has a conn.

Then there's the inconsistency assertions at the end of error paths.   Ignoring errors needs to be removed, and we could keep that change to not confuse commit errors with inconsistency if we fixed it up.

I think that's everything?   It'll be a lot clearer once the themes are separated out.

versity-github:
Build 117 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/117/ git: 5df88ba035d0fb782102a8f7fd55dbc78bd4affd

versity-github:
Build 118 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/118/ git: 7f126da4660b2686a51de79a67f8c74600d54ad0

versity-github:
Build 120 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/120/ git: 24d53324bb25aef705029a5ddc38bb70bd18df3f

kmod/src/server.c
```diff
@@ -3881,6 +3883,7 @@ static void scoutfs_server_worker(struct work_struct *work)
 	scoutfs_alloc_init(&server->alloc,
 			   &super->server_meta_avail[server->other_ind ^ 1],
 			   &super->server_meta_freed[server->other_ind ^ 1]);
+	alloc_init=true;
```
zabbo, line 3897: Spaces around = (and a trivial thoes typo in the message that I can't comment on specifically \o/)

kmod/src/server.c
```diff
@@ -2533,9 +2534,19 @@ static int server_commit_log_merge(struct super_block *sb,
 	if (ret < 0)
 		scoutfs_err(sb, "error %d committing log merge: %s", ret, err_str);
 
-	ret = scoutfs_server_apply_commit(sb, ret);
 	BUG_ON(ret < 0); /* inconsistent */
 
+	if (err < 0) {
```
zabbo, line 2539: This all is pretty close, but still a bit off.   The big problem is that _apply_commit() is like an unlock.  We always have to call it after having called hold_commit or the commit_rwsem will still be held and server tasks will hang.  The other functional requirement is that we generally don't want to use BUG_ON while holding locks for the same reason, it kills the tasks and leaves locks held and the rest of the world hangs.  Then there's the stylistic preference of using err sparingly and when only necessary.   So putting all that together, I think we want: add a deleted bool initialized to false and set it to true after the initial deletion succeeds.   only use err to store the result of the apply_commit.   Leave the bug_on and apply where they are, but change the BUG conditino to ret < 0 && deleted.   Then after the bug set ret to err if ret was 0 so that it's sent as the net_response.   I think that's a minimal change that only triggers the bug if we're inconsistent.

kmod/src/server.c
```diff
@@ -4095,3 +4095,11 @@ void scoutfs_server_destroy(struct super_block *sb)
 		sbi->server_info = NULL;
 	}
 }
+
+void scoutfs_server_net_shutdown(struct super_block *sb)
```
zabbo, line 4099: It's not safe to reference the server-> fields here like it.. kind of is in the client fields.   The client conn is a long lived container that is up for the duration of the mount.   If we're calling umount_begin then we've built up the mount and haven't torn it down yet -- we know the conn is still around.   The server is on a totally different life cycle.  It's built up and torn down as quorum is elected leader and the server decides to shutdown for whatever reason it wants.   forced unmont can race with the server shutting down and load a valid server_info pointer that is being actively freed while the server is shutting down.  Happily, I don't think we have to actively shutdown the server's listening socket.  And if we did, we'd just stop the server rather than trying to reach in and shutdown the conn.   Let's remove the server half of this and see if the system cooperates.

kmod/src/btree.c
```diff
@@ -1234,9 +1234,6 @@ static int btree_walk(struct super_block *sb,
 		return -EINVAL;
 
 	/* all ops come through walk and walk calls all reads */
```
zabbo, line 1236: Remove the comment along with the test too :)

tests/funcs/filter.sh
```diff
@@ -72,6 +72,9 @@ t_filter_dmesg()
 	re="$re|scoutfs .* error reading quorum block"
 	re="$re|scoutfs .* error .* writing quorum block"
 	re="$re|scoutfs .* error .* while checking to delete inode"
+	re="$re|scoutfs .* error .* freeing merged btree blocks.*.looping commit del.*upd freeing item"
```
zabbo, line 75: hmm?   regexes don't implement multiple matches like this?   I'm so confused.  It looks like there are multiple messages here that should be multiple lines.   And that's the intent of the style.. make it easier to add and remove messages as their sources are.  Rather than trying to maintain some dense multiple regex like (a|b|c).  Decent regex runtimes will compile it down into an optimal form anyway, we don't have to worry about redundant input descriptions.

zabbo (changes requested):
These broken out patches are great, just a few more tweaks and I think we're there.

versity-github:
Build 121 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/121/ git: 6e2e37117317f9b1cf30cc245cb46ec48b6d41bc

versity-github:
Build 122 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/122/ git: d3ea4a28da15fbcd5d5f0137a208a4cb2be56bf1

bgly:
Patch is ready for review. It was able to complete runs the entire weekend without any issues. Runs being looping around the forced unmount test cases the entire weekend.

versity-github:
Build 123 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/123/ git: 38ee2defd567818d6aa7c879eecf161c13c750b6

zabbo (approved):

PR #57, opened 2021-11-05
Update README.md

zabbo:
Remove a bunch of old language from the README.  We're no longer in the
early days of the open release so we can remove all the alpha quality
language.   And the system has grown sufficiently that the repo README
isn't a great place for a small getting started doc.  There just isn't
room to do the subject justice.   If we need such a thing for the
project we'll put it as a first order doc in the repo that'd be
distributed along with everything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,135 +1,24 @@
 # Introduction
 
-scoutfs is a clustered in-kernel Linux filesystem designed and built
-from the ground up to support large archival systems.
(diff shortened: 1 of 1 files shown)


bgly (approved):

versity-github:
Build 119 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/119/ git: 465e5ee769a68834be40b027b56f0362d6446a47

PR #60, opened 2021-11-15
Fix block-stale-read test case

bgly:
The current test case attempts to create a state to read
by calling setattr and getattr in attempt to force block
cache reads. It so happens that this does not always force
cache block reads, which in rare cases causes this test case
to fail.

The new test case removes all the extra bouncing around of mount
points and we just directly call scoutfs df which will walk
everyone's allocators to summarize the block counts, which is
guaranteed to exist. Therefore, we do not have to create any sort
of state prior to trying to force a read.

Signed-off-by: Bryant G. Duffy-Ly <bduffyly@versity.com>


versity-github:
Build 129 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/129/ git: 6e4ee12df58fe2c48d2213d1df3b8fddd5a4f8d6

bgly:
Btw test case will fail because I am changing the test case which will make it so output differs.

tests/golden/block-stale-reads
```diff
@@ -1,52 +1,3 @@
...
-user.xat="9"
-
-counter block_cache_remove_stale changed
-counter block_cache_remove_stale changed
-# file: /mnt/test/test/block-stale-reads/file
-user.xat="10"
-
-counter block_cache_remove_stale changed
-counter block_cache_remove_stale changed
+counter block_cache_remove_stale changed
```
zabbo, line 3: The test only failed 'cause this line had its newline removed in the golden output but the test still echoed the line with a newline.

tests/tests/block-stale-reads.sh
```diff
@@ -6,56 +6,21 @@
...
-		nr=0
-	fi
-
-	echo $nr
-}
-
-GETFATTR="getfattr --absolute-names"
-SETFATTR="setfattr"
-
 echo "== create shared test file"
```
zabbo, line 9: These little status messages should be updated to say what the new test is doing.   I'd get rid of most of them and just echo that it's using df to issue block reads that trigger the stale invalidation and retry.

tests/tests/block-stale-reads.sh
```diff
@@ -6,56 +6,21 @@
...
-# their trigger fired prematurely by message handling btree calls while
-# working with the t_ helpers long before we work with the xattrs.  But
-# the block cache stale retry path is still being exercised.
-#
+# Trigger retries in the block cache by calling scoutfs df
+# which in turn will call scoutfs_ioctl_alloc_detail. This
+# is guaranteed to exist, which will force block cache reads.
+
 echo "== set and get xattrs between mount pairs while retrying"
 set_nr=0
```
zabbo, line 17: The set and get names were from setattr and getattr, which we're not doing anymore.   This should be just 'nr' which the tests tend to use as the mount nr.  It matches using M0 as the df path later.

tests/tests/block-stale-reads.sh
```diff
@@ -6,56 +6,21 @@
...
-	t_counter_diff_changed block_cache_remove_stale $old_get $get_nr
+scoutfs df -p "$T_M0" > /dev/null
 
-	set_nr="$get_nr"
-	get_nr=$(inc_wrap_fs_nr $set_nr)
-done
+t_counter_diff_changed block_cache_remove_stale $old_set $set_nr
 
-t_pass
+t_pass
```
zabbo, line 26: Let's not add diff noise by removing newlines (and it git yells about missing newlines so that lines can later be added without making it look like we modified the existing last line)

zabbo (changes requested):
In addition to the review comments, we probably don't need the full failed test output.  It's a bit noisy for commit messages.   Just describing that the counters wouldn't change is fine.

versity-github:
Build 133 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/133/ git: c755141153ce1bc296b155e0b6634b70b940e407

versity-github:
Build 134 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/134/ git: 2bdfc20cd15316eb531309cfea0e2378e6882adc

versity-github:
Build 138 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/138/ git: 8b8f8d2926044bfecdc4a3e6e83b89f30fc1770e

tests/tests/block-stale-reads.sh
```diff
@@ -5,57 +5,20 @@
...
+# is guaranteed to exist, which will force block cache reads.
 
-t_require_commands touch setfattr getfattr
+echo "== Issue scoutfs df to force block reads to trigger stale invalidation/retry"
+nr=0
 
-inc_wrap_fs_nr()
-{
-	local nr="$(($1 + 1))"
+old_set=$(t_counter block_cache_remove_stale $nr)
```
zabbo, line 15: We don't need the _set suffix from the previous get/set test.    Just old, I guess?

tests/tests/block-stale-reads.sh
```diff
@@ -5,57 +5,20 @@
...
+t_trigger_arm_silent block_remove_stale $nr
 
-	if [ "$nr" == "$T_NR_MOUNTS" ]; then
-		nr=0
-	fi
+scoutfs df -p "$T_M0" > /dev/null
 
-	echo $nr
-}
+echo ""
```
zabbo, line 20: What's this extra echo?

zabbo (changes requested):
Looks great, just tiny stuff and we're good.

versity-github:
Build 142 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/142/ git: 3ae0ebd0d8554c7470fc262cfe85b66065d69c64

versity-github:
Build 143 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/143/ git: 3ae0ebd0d8554c7470fc262cfe85b66065d69c64

zabbo (approved):

PR #70, opened 2022-02-02
Silence resent log merge commit error

zabbo:
The server's log merge complete request handler was considering the
absence of the client's original request as a failure.  Unfortunately,
this case is possible if a previous server successfully completed the
client's request but the response was lost because it stopped for
whatever reason.

The failure was being logged as a hard error to the console which was
causing tests to occasionally fail during server failover that hit just
as the log merge completion was being processed.

The error was being sent to the client as a response, we just need to
silence the message for these expected but rare errors.

We also fix the related case where the server printed the even more
harsh WARN_ON if there was a next original request but it wasn't the one
we expected to find from our requesting client.

### DIFF
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -2455,15 +2455,14 @@ static int server_commit_log_merge(struct super_block *sb,
 	}
 
 	/* find the completion's original saved request */
-	ret = next_log_merge_item(sb, &super->log_merge,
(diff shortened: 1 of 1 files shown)


versity-github:
Build 178 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/178/ git: 730a84af92a4105f7f80b21b1b0de282aa0f4f59

bgly (approved):

PR #73, opened 2022-02-10
Fix calls that check for quorum leader

bgly:
Currently the calls do not check if the file exists; which in the
client-unmount-recovery unit test where in some cases the is_leader file
might not exist based on timing then the unit test fails. It would be better
to add the extra check in the calls to t_server_nr and t_first_client_nr
to make sure the file exists prior to trying to cat for is_leader.

Signed-off-by: Bryant G. Duffy-Ly <bduffyly@versity.com>

### DIFF
--- a/tests/funcs/fs.sh
+++ b/tests/funcs/fs.sh
@@ -83,7 +83,7 @@ t_fs_nrs()
 t_server_nr()
 {
 	for i in $(t_fs_nrs); do
-		if [ "$(cat $(t_sysfs_path $i)/quorum/is_leader)" == "1" ]; then
+		if [ "$(cat $(t_sysfs_path $i)/quorum/is_leader 2>&1)" == "1" ]; then
 			echo $i
 			return
 		fi
@@ -101,7 +101,7 @@ t_server_nr()
 t_first_client_nr()
 {
 	for i in $(t_fs_nrs); do
-		if [ "$(cat $(t_sysfs_path $i)/quorum/is_leader)" == "0" ]; then
+		if [ "$(cat $(t_sysfs_path $i)/quorum/is_leader 2>&1)" == "0" ]; then
 			echo $i
 			return
 		fi


tests/funcs/fs.sh
```diff
@@ -101,7 +101,7 @@ t_server_nr()
 t_first_client_nr()
 {
 	for i in $(t_fs_nrs); do
-		if [ "$(cat $(t_sysfs_path $i)/quorum/is_leader)" == "0" ]; then
+		if if [[ -f "$(t_sysfs_path $i)/quorum/is_leader" ]] && [[ "$(cat $(t_sysfs_path $i)/quorum/is_leader)" == "0" ]]; then
```
zabbo, line 104: `if if`? :)

This doesn't remove the race, it just makes it tighter.  The file can be removed after we test it but before we try and cat it.   The fix is to just add `2>&1` to the cats so that it compares the error message to 1 and fails rather than it ending up in the output and making the test fail.  (which is what I assume is happening :)).

zabbo (changes requested):
Nice find, but we need to fix the fix just a bit.

versity-github:
Build 186 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/186/ git: 7feffe78ca1ee5c730bd3487decef707bab7716d

versity-github:
Build 187 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/187/ git: 4c7dd4577722e6fd761138fd853aa33333f91f04

zabbo (changes requested):
Hmm, well now the commit message doesn't describe the change.   And poking around a bit more, I'm not sure I understand the failure.

The waits in the test should be stabilizing the mounts.   How is t_server_nr seeing unstable is_leader files?   (And if it doesn't get a server nr its going to fail horribly, just making the cats not add output isn't going to fix the test if there isn't a server when it needs one.)

PR #84, opened 2022-04-04
Fix getxattr with large values giving EINVAL

zabbo:
The change to only allocate a buffer for the first xattr item with
kmalloc instead of the entire logical xattr payload with vmalloc
included a regression for getting large xattrs.

getxattr used to copy the entire payload into the large vmalloc so it
could unlock just after get_next_xattr.   The change to only getting the
first item buffer added a call to copy from the rest of the items but
those copies weren't covered by the locks.  This would often work
because the lock pointer still pointed to a valid lock.  But if the lock
was invalidated then the mode would no longer be compatible and
_item_lookup would return EINVAL.

The fix is to extend xattr_rwsem and cluster lock coverage to the rest
fo the function body, which includes the value item copies.  This also
makes getxattr's lock coverage consistent with setxattr and listxattr
which might reduce the risk of similar mistakes in the future.

### DIFF
--- a/kmod/src/xattr.c
+++ b/kmod/src/xattr.c
@@ -488,28 +488,28 @@ ssize_t scoutfs_getxattr(struct dentry *dentry, const char *name, void *buffer,
 
 	ret = get_next_xattr(inode, &key, xat, xat_bytes, name, name_len, 0, 0, lck);
 
-	up_read(&si->xattr_rwsem);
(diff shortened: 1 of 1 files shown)


bgly (approved):

versity-github:
Build 222 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/222/ git: 65654ee7c00bf84e8a4fd9faf32f0459a4335d7e

PR #90, opened 2022-06-08
Reclaim log_trees alloc roots in multiple commits

zabbo:
Client log_trees allocator btrees can build up quite a number of
extents.  In the right circumstances fragmented extents can have to
dirty a large number of paths to leaf blocks in the core allocator
btrees.  It might not be possible to dirty all the blocks necessary to
move all the extents in one commit.

This reworks the extent motion so that it can be performed in multiple
commits if the meta allocator for the commit runs out while it is moving
extents.  It's a minimal fix with as little disruption to the ordering
of commits and locking as possible.  It simply bubbles up an error when
the allocators run out and retries functions that can already be retried
in other circumstances.

### DIFF
--- a/kmod/src/alloc.c
+++ b/kmod/src/alloc.c
@@ -85,6 +85,21 @@ static u64 smallest_order_length(u64 len)
 }
 
 /*
+ * An extent modification dirties three distinct leaves of an allocator
+ * btree as it adds and removes the blkno and size sorted items for the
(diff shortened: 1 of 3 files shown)


versity-github:
Build 230 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/230/ git: 0d4bf83da3dde47fc989c06e60355d4c9d62ce92

PR #91, opened 2022-06-14
Set sk_allocation on TCP sockets

zabbo:
We were setting sk_allocation on the quorum UDP sockets to prevent
entering reclaim while using sockets but we missed setting it on the
regular messaging TCP sockets.   This could create deadlocks where the
sending socket could enter scoutfs reclaim and wait for server messages
while holding the socket lock, preventing the receive thread from
receiving messages while it blocked on the socket lock.

The fix is to prevent entering the FS to reclaim during socket
allocations.

### DIFF
--- a/kmod/src/net.c
+++ b/kmod/src/net.c
@@ -991,6 +991,8 @@ static void scoutfs_net_listen_worker(struct work_struct *work)
 		if (ret < 0)
 			break;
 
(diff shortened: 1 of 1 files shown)


versity-github:
Build 231 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/231/ git: ae55fa3153d4aef94a13802bfae979311574657c

PR #92, opened 2022-06-17
Protect get_log_trees corruption with assertion

zabbo:
Like a lot of places in the server, get_log_trees() doesn't have the
tools in needs to safely unwind partial changes in the face of an error.

In the worst case, it can have moved extents from the mount's log_trees
item into the server's main data allocator.  The dirty data allocator
reference is in the super block so it can be written later.   The dirty
log_trees reference is on stack, though, so it will be thrown away on
error.  This ends up duplicating extents in the persistent structures
because they're written in the new dirty allocator but still remain in
the unwritten source log_trees allocator.

This change makes it harder for that to happen.   It dirties the
log_trees item and always tries to update so that the dirty blocks are
consistent if they're later written out.  If we do get an error updating
the item we throw an assertion.   It's not great, but it matches other
similar circumstances in other parts of the server.

### DIFF
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -1267,6 +1267,7 @@ static int server_get_log_trees(struct super_block *sb,
 	char *err_str = NULL;
 	u64 nr;
 	int ret;
+	int err;
 
 	if (arg_len != 0) {
 		ret = -EINVAL;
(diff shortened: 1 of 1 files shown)


versity-github:
Build 232 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/232/ git: dd6f59f480aefc257184e6a5cd698d0633624dea

versity-github:
Build 233 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/233/ git: 31e474c5fa7077bbb99719b16562f54f580244d9

PR #113, opened 2023-01-10
Fix move_blocks loop exit conditions

zabbo:
The move_blocks ioctl intends to only move extents whose bytes fall inside i_size.  This is easy except for a final extent that straddles an i_size that isn't aligned to 4K data blocks.

The code that either checked for an extent being entirely past i_size or for limiting the number of blocks to move by i_size clumsily compared i_size offsets in bytes with extent counts in 4KB blocks.  In just the right circumstances, probably with the help of a byte length to move that is much larger than i_size, the length calculation could result in trying to move 0 blocks.  Once this hit the loop would keep finding that extent and calculating 0 blocks to move and would be stuck.

We fix this by clamping the count of blocks in extents to move in terms of byte offsets at the start of the loop.  This gets rid of the extra size checks and byte offset use in the loop.  We also add a sanity check to make sure that we can't get stuck if, say, corruption resulted in an otherwise impossible zero length extent.

### DIFF
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -1192,9 +1192,9 @@ static void truncate_inode_pages_extent(struct inode *inode, u64 start, u64 len)
  * explained above the move_blocks ioctl argument structure definition.
  *
  * The caller has processed the ioctl args and performed the most basic
- * inode checks, but we perform more detailed inode checks once we have
- * the inode lock and refreshed inodes.  Our job is to safely lock the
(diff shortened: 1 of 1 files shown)


versity-github:
Build 261 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/261/ git: a23e7478a00d0d49600c98e5832f806464872017

PR #114, opened 2023-01-17
Allow replaying srch file rotation

zabbo:
When a client no longer needs to append to a srch file, for whatever reason, we move the reference from the log_trees item into a specific srch file btree item in the server's srch file tracking btree.

Zeroing the log_trees item and inserting the server's btree item are done in a server commit and should be written atomically.

But commit_log_trees had an error handling case that could leave the newly inserted item dirty in memory without zeroing the srch file reference in the existing log_trees item.  Future attempts to rotate the file reference, perhaps by retrying the commit or by reclaiming the client's rid, would get EEXIST and fail.

This fixes the error handling path to ensure that we'll keep the dirty srch file btree and log_trees item in sync.  The desynced items can still exist in the world so we'll tolerate getting EEXIST on insertion. After enough time has passed, or if repair zeroed the duplicate reference, we could remove this special case from insertion.

### DIFF
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -1586,6 +1586,13 @@ static int server_commit_log_trees(struct super_block *sb,
 	if (ret < 0 || committed)
 		goto unlock;
 
+	/* make sure _update succeeds before we modify srch items */
+	ret = scoutfs_btree_dirty(sb, &server->alloc, &server->wri, &super->logs_root, &key);
+	if (ret < 0) {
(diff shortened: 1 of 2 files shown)


versity-github:
Build 264 http://jenkins-prod.vpn.versity.com:8080/job/scoutfs-run-pr/264/ git: 2e2ccb6f61cd8cb3d1660b79dbb08b51a890a39d

PR #123, opened 2023-05-22
el8 support

aversecat:
~~WIP~~

### DIFF
--- a/kmod/src/Makefile
+++ b/kmod/src/Makefile
@@ -25,6 +25,7 @@ scoutfs-y +=			\
 	inode.o			\
 	ioctl.o			\
 	item.o			\
+	kernelcompat.o		\
 	lock.o			\
 	lock_server.o		\
 	msg.o			\
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -204,3 +204,12 @@ endif
 ifneq (,$(shell grep 'define register_hotcpu_notifier' include/linux/cpu.h))
 ccflags-y += -DKC_CPU_NOTIFIER
 endif
+
+#
+# v3.14-rc8-130-gccad2365668f
+#
+# generic_file_buffered_write is removed, backport it
+#
+ifneq (,$(shell grep 'extern ssize_t generic_file_buffered_write' include/linux/fs.h))
+ccflags-y += -DKC_GENERIC_FILE_BUFFERED_WRITE=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -53,3 +53,13 @@ endif
 ifneq (,$(shell grep 'posix_acl_valid.*user_ns,' include/linux/posix_acl.h))
 ccflags-y += -DKC_POSIX_ACL_VALID_USER_NS
 endif
+
+#
+# v5.3-12296-g6d2052d188d9
+#
+# The RBCOMPUTE function is now passed an extra flag, and should return a bool
+# to indicate whether the propagated callback should stop or not.
+#
+ifneq (,$(shell grep 'static inline bool RBNAME.*_compute_max' include/linux/rbtree_augmented.h))
+ccflags-y += -DKC_RB_TREE_AUGMENTED_COMPUTE_MAX
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -130,3 +130,12 @@ endif
 ifneq (,$(shell grep 'bi_status' include/linux/blk_types.h))
 ccflags-y += -DKC_BIO_BI_STATUS
 endif
+
+#
+# v3.11-8765-ga0b02131c5fc
+#
+# Remove the old ->shrink() API, ->{scan,count}_objects is preferred.
+#
+ifneq (,$(shell grep '(*shrink)' include/linux/shrinker.h))
+ccflags-y += -DKC_SHRINKER_SHRINK
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -185,3 +185,12 @@ endif
 ifneq (,$(shell grep 'struct iov_iter.*msg_iter' include/linux/socket.h))
 ccflags-y += -DKC_MSGHDR_STRUCT_IOV_ITER=1
 endif
+
+#
+# v4.17-rc6-7-g95582b008388
+#
+# Kernel has current_time(inode) to uniformly retreive timespec in the right unit
+#
+ifneq (,$(shell grep 'extern struct timespec64 current_time' include/linux/fs.h))
+ccflags-y += -DKC_CURRENT_TIME_INODE=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -92,3 +92,12 @@ endif
 ifneq (,$(shell grep -s 'define _LINUX_IVERSION_H' include/linux/iversion.h))
 ccflags-y += -DKC_NEED_LINUX_IVERSION_H=1
 endif
+
+# v4.11-12447-g104b4e5139fe
+#
+# Renamed __percpu_counter_add to percpu_counter_add_batch to clarify
+# that the __ wasn't less safe, just took an extra parameter.
+#
+ifneq (,$(shell grep 'percpu_counter_add_batch' include/linux/percpu_counter.h))
+ccflags-y += -DKC_PERCPU_COUNTER_ADD_BATCH
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -50,7 +50,7 @@ endif
 #
 # Added user_ns argument to posix_acl_valid
 #
-ifneq (,$(shell grep 'posix_acl_valid.*user_ns,' include/linux/posix_acl.h))
+ifneq (,$(shell grep 'posix_acl_valid.*user_namespace' include/linux/posix_acl.h))
 ccflags-y += -DKC_POSIX_ACL_VALID_USER_NS
 endif
 
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -101,3 +101,12 @@ endif
 ifneq (,$(shell grep 'percpu_counter_add_batch' include/linux/percpu_counter.h))
 ccflags-y += -DKC_PERCPU_COUNTER_ADD_BATCH
 endif
+
+#
+# v4.11-4550-g7dea19f9ee63
+#
+# Introduced memalloc_nofs_{save,restore} preferred instead of _noio_.
+#
+ifneq (,$(shell grep 'memalloc_nofs_save' include/linux/sched/mm.h))
+ccflags-y += -DKC_MEMALLOC_NOFS_SAVE
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -213,3 +213,15 @@ endif
 ifneq (,$(shell grep 'extern ssize_t generic_file_buffered_write' include/linux/fs.h))
 ccflags-y += -DKC_GENERIC_FILE_BUFFERED_WRITE=1
 endif
+
+#
+# v5.7-438-g8151b4c8bee4
+#
+# struct address_space_operations switches away from .readpages to .readahead
+#
+# RHEL has backported this feature all the way to RHEL8, as part of RHEL_KABI,
+# which means we need to detect this very precisely
+#
+ifneq (,$(shell grep 'readahead.*struct readahead_control' include/linux/fs.h))
+ccflags-y += -DKC_FILE_AOPS_READAHEAD
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -194,3 +194,13 @@ endif
 ifneq (,$(shell grep 'extern struct timespec64 current_time' include/linux/fs.h))
 ccflags-y += -DKC_CURRENT_TIME_INODE=1
 endif
+
+#
+# v4.9-12228-g530e9b76ae8f
+#
+# register_cpu_notifier and family were all removed and to be
+# replaced with cpuhp_* API calls.
+#
+ifneq (,$(shell grep 'define register_hotcpu_notifier' include/linux/cpu.h))
+ccflags-y += -DKC_CPU_NOTIFIER
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -65,6 +65,16 @@ ccflags-y += -DKC_RB_TREE_AUGMENTED_COMPUTE_MAX
 endif
 
 #
+# v3.13-25-g37bc15392a23
+#
+# Renames posix_acl_create to __posix_acl_create and provide some
+# new interfaces for creating ACLs
+#
+ifneq (,$(shell grep '__posix_acl_create' include/linux/posix_acl.h))
+ccflags-y += -DKC___POSIX_ACL_CREATE
+endif
+
+#
 # v4.8-rc1-29-g31051c85b5e2
 #
 # inode_change_ok() removed - replace with setattr_prepare()
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -65,6 +65,15 @@ ccflags-y += -DKC_RB_TREE_AUGMENTED_COMPUTE_MAX
 endif
 
 #
+# v4.8-rc1-29-g31051c85b5e2
+#
+# inode_change_ok() removed - replace with setattr_prepare()
+#
+ifneq (,$(shell grep 'extern int setattr_prepare' include/linux/fs.h))
+ccflags-y += -DKC_SETATTR_PREPARE
+endif
+
+#
 # v4.15-rc3-4-gae5e165d855d
 #
 # linux/iversion.h needs to manually be included for code that
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -63,3 +63,13 @@ endif
 ifneq (,$(shell grep 'static inline bool RBNAME.*_compute_max' include/linux/rbtree_augmented.h))
 ccflags-y += -DKC_RB_TREE_AUGMENTED_COMPUTE_MAX
 endif
+
+#
+# v4.15-rc3-4-gae5e165d855d
+#
+# linux/iversion.h needs to manually be included for code that
+# manipulates this field.
+#
+ifneq (,$(shell grep -s 'define _LINUX_IVERSION_H' include/linux/iversion.h))
+ccflags-y += -DKC_NEED_LINUX_IVERSION_H=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -27,6 +27,16 @@ ccflags-y += -DKC_DIR_EMIT_DOTS
 endif
 
 #
+# v3.18-rc2-19-gb5ae6b15bd73
+# 
+# Folds d_materialise_unique into d_splice_alias. Note reversal
+# of arguments (Also note Documentation/filesystems/porting.rst)
+#
+ifneq (,$(shell grep 'd_materialise_unique' include/linux/dcache.h))
+ccflags-y += -DKC_D_MATERIALISE_UNIQUE=1
+endif
+
+#
 # RHEL extended the fop struct so to use it we have to set
 # a flag to indicate that the struct is large enough and
 # contains the pointer.
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -167,3 +167,12 @@ endif
 ifneq (,$(shell grep 'kernel_getsockname.*,$$' include/linux/net.h))
 ccflags-y += -DKC_KERNEL_GETSOCKNAME_ADDRLEN=1
 endif
+
+#
+# v4.1-rc1-410-geeb1bd5c40ed
+#
+# Adds a struct net parameter to sock_create_kern
+#
+ifneq (,$(shell grep 'sock_create_kern.*struct net' include/linux/net.h))
+ccflags-y += -DKC_SOCK_CREATE_KERN_NET=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -110,3 +110,23 @@ endif
 ifneq (,$(shell grep 'memalloc_nofs_save' include/linux/sched/mm.h))
 ccflags-y += -DKC_MEMALLOC_NOFS_SAVE
 endif
+
+#
+# v4.7-12414-g1eff9d322a44
+#
+# Renamed bi_rw to bi_opf to force old code to catch up.  We use it as a
+# single switch between old and new bio structures.
+#
+ifneq (,$(shell grep 'bi_opf' include/linux/blk_types.h))
+ccflags-y += -DKC_BIO_BI_OPF
+endif
+
+#
+# v4.12-rc2-201-g4e4cbee93d56
+#
+# Moves to bi_status BLK_STS_ API instead of having a mix of error
+# end_io args or bi_error.
+#
+ifneq (,$(shell grep 'bi_status' include/linux/blk_types.h))
+ccflags-y += -DKC_BIO_BI_STATUS
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -225,3 +225,26 @@ endif
 ifneq (,$(shell grep 'readahead.*struct readahead_control' include/linux/fs.h))
 ccflags-y += -DKC_FILE_AOPS_READAHEAD
 endif
+
+#
+# v4.0-rc7-1743-g8436318205b9
+#
+# .aio_read and .aio_write no longer exist. All reads and writes now use the
+# .read_iter and .write_iter methods, or must implement .read and .write (which
+# we don't).
+#
+ifneq (,$(shell grep 'ssize_t.*aio_read' include/linux/fs.h))
+ccflags-y += -DKC_LINUX_HAVE_FOP_AIO_READ=1
+endif
+
+#
+# rhel7 has a custom inode_operations_wrapper struct that is discarded
+# entirely in favor of upstream structure since rhel8.
+#
+ifneq (,$(shell grep 'void.*follow_link.*struct dentry' include/linux/fs.h))
+ccflags-y += -DKC_LINUX_HAVE_RHEL_IOPS_WRAPPER=1
+endif
+
+ifneq (,$(shell grep 'size_t.*ki_left;' include/linux/aio.h))
+ccflags-y += -DKC_LINUX_AIO_KI_LEFT=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -139,3 +139,13 @@ endif
 ifneq (,$(shell grep '(*shrink)' include/linux/shrinker.h))
 ccflags-y += -DKC_SHRINKER_SHRINK
 endif
+
+#
+# v3.19-4777-g6bec00352861
+#
+# backing_dev_info is removed from address_space. Instead we need to use
+# inode_to_bdi() inline from <backing-dev.h>.
+#
+ifneq (,$(shell grep 'struct backing_dev_info.*backing_dev_info' include/linux/fs.h))
+ccflags-y += -DKC_LINUX_BACKING_DEV_INFO=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -248,3 +248,13 @@ endif
 ifneq (,$(shell grep 'size_t.*ki_left;' include/linux/aio.h))
 ccflags-y += -DKC_LINUX_AIO_KI_LEFT=1
 endif
+
+#
+# v4.4-rc4-4-g98e9cb5711c6
+#
+# Introduces a new xattr_handler .name member that can be used to match the
+# entire field, instead of just a prefix. For these kernels, we must use
+# the new .name field instead.
+ifneq (,$(shell grep 'static inline const char .xattr_prefix' include/linux/xattr.h))
+ccflags-y += -DKC_XATTR_HANDLER_NAME=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -176,3 +176,12 @@ endif
 ifneq (,$(shell grep 'sock_create_kern.*struct net' include/linux/net.h))
 ccflags-y += -DKC_SOCK_CREATE_KERN_NET=1
 endif
+
+#
+# v3.18-rc6-1619-gc0371da6047a
+#
+# iov_iter is now part of struct msghdr
+#
+ifneq (,$(shell grep 'struct iov_iter.*msg_iter' include/linux/socket.h))
+ccflags-y += -DKC_MSGHDR_STRUCT_IOV_ITER=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -149,3 +149,12 @@ endif
 ifneq (,$(shell grep 'struct backing_dev_info.*backing_dev_info' include/linux/fs.h))
 ccflags-y += -DKC_LINUX_BACKING_DEV_INFO=1
 endif
+
+#
+# v4.3-9290-ge409de992e3e
+#
+# xattr handlers are now passed a struct that contains `flags`
+#
+ifneq (,$(shell grep 'int...get..const struct xattr_handler.*struct dentry.*dentry,' include/linux/xattr.h))
+ccflags-y += -DKC_XATTR_STRUCT_XATTR_HANDLER=1
+endif
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -158,3 +158,12 @@ endif
 ifneq (,$(shell grep 'int...get..const struct xattr_handler.*struct dentry.*dentry,' include/linux/xattr.h))
 ccflags-y += -DKC_XATTR_STRUCT_XATTR_HANDLER=1
 endif
+
+#
+# v4.16-rc1-1-g9b2c45d479d0
+#
+# kernel_getsockname() and kernel_getpeername dropped addrlen arg
+#
+ifneq (,$(shell grep 'kernel_getsockname.*,$$' include/linux/net.h))
+ccflags-y += -DKC_KERNEL_GETSOCKNAME_ADDRLEN=1
+endif
--- a/kmod/src/acl.c
+++ b/kmod/src/acl.c
@@ -218,10 +218,17 @@ int scoutfs_set_acl(struct inode *inode, struct posix_acl *acl, int type)
 	scoutfs_unlock(sb, lock, SCOUTFS_LOCK_WRITE);
 	return ret;
 }
-
+#ifdef KC_XATTR_STRUCT_XATTR_HANDLER
+int scoutfs_acl_get_xattr(const struct xattr_handler *handler, struct dentry *dentry,
+			  struct inode *inode, const char *name, void *value,
+			  size_t size)
+{
+	int type = handler->flags;
+#else
 int scoutfs_acl_get_xattr(struct dentry *dentry, const char *name, void *value, size_t size,
 			  int type)
 {
+#endif
 	struct posix_acl *acl;
 	int ret = 0;
 
@@ -240,9 +247,17 @@ int scoutfs_acl_get_xattr(struct dentry *dentry, const char *name, void *value, 
 	return ret;
 }
 
+#ifdef KC_XATTR_STRUCT_XATTR_HANDLER
+int scoutfs_acl_set_xattr(const struct xattr_handler *handler, struct dentry *dentry,
+			  struct inode *inode, const char *name, const void *value,
+			  size_t size, int flags)
+{
+	int type = handler->flags;
+#else
 int scoutfs_acl_set_xattr(struct dentry *dentry, const char *name, const void *value, size_t size,
 			  int flags, int type)
 {
+#endif
 	struct posix_acl *acl = NULL;
 	int ret;
 
--- a/kmod/src/acl.c
+++ b/kmod/src/acl.c
@@ -69,12 +69,14 @@ struct posix_acl *scoutfs_get_acl_locked(struct inode *inode, int type, struct s
 	char *name;
 	int ret;
 
+#ifndef KC___POSIX_ACL_CREATE
 	if (!IS_POSIXACL(inode))
 		return NULL;
 
 	acl = get_cached_acl(inode, type);
 	if (acl != ACL_NOT_CACHED)
 		return acl;
+#endif
 
 	ret = acl_xattr_name_len(type, &name, NULL);
 	if (ret < 0)
@@ -96,9 +98,11 @@ struct posix_acl *scoutfs_get_acl_locked(struct inode *inode, int type, struct s
 		acl = ERR_PTR(ret);
 	}
 
+#ifndef KC___POSIX_ACL_CREATE
 	/* can set null negative cache */
 	if (!IS_ERR(acl))
 		set_cached_acl(inode, type, acl);
+#endif
 
 	kfree(value);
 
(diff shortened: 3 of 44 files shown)


versity-github:
Build 12 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/12/ git: 8ad9e56df251e4c043d6d00a8b70a58cf88cc7f8

kmod/src/Makefile.kernelcompat
```diff
@@ -40,8 +40,15 @@ endif
 #
 # Added user_ns argument to posix_acl_valid
 #
-ifneq (,$(shell grep 'posix_acl_valid.*user_ns,' include/linux/posix_acl.h))
-ccflags-y += -DKC_POSIX_ACL_VALID_USER_NS
+ifneq (,$(shell grep 'posix_acl_valid.*user_namespace' include/linux/posix_acl.h))
```
zabbo, line 43: It's only the definition that changed, the code still calls it user_ns so let's not change the KC define.  Makes it a tiny patch on its own.

kmod/src/item.c
```diff
@@ -2277,7 +2277,7 @@ int scoutfs_item_write_dirty(struct super_block *sb)
 		ret = -ENOMEM;
 		goto out;
 	}
-	list_add(&page->list, &pages);
+	list_add(&page->lru, &pages);
```
zabbo, line 2280: This list->lru change should be in its own patch.

kmod/src/item.c
```diff
@@ -2530,27 +2530,42 @@ void scoutfs_item_invalidate(struct super_block *sb, struct scoutfs_key *start,
...
+{
+	struct item_cache_info *cinf = container_of(shrink,
+						    struct item_cache_info,
+						    shrinker);
+	struct super_block *sb = cinf->sb;
+
+	scoutfs_inc_counter(sb, item_cache_scan_objects);
+
+	//FIXME just picked ->lru_pages here being dumb
+	return min_t(u64, (u64)(&cinf->lru_pages) * SCOUTFS_BLOCK_LG_PAGES_PER,
```
zabbo, line 2544: The * LG_PAGES_PER only makes sense for the block cache.  Remove it from the others.

kmod/src/kernelcompat.h
```diff
@@ -174,33 +174,32 @@ do {					\
...
+#define KC_DEFINE_SHRINKER(name) struct				\
 	{							\
 		struct kc_shrinker_funcs shrinker_funcs;	\
-		struct shinker name;				\
+		struct shrinker (name);				\
 	}
-#define KC_SHRINKER_FUNCS(shrinker) \
-	((void *)((long)(shrink) - sizeof(struct kc_shrinker_funcs)))
-#define KC_INIT_SHRINKER_FUNCS(type, name, shrink, count, scan) do {				\
-	BUILD_BUG_ON(offsetof(cont, shrink_funcs) + sizeof(struct kc_shrinker_funcs)) !=	\
```
zabbo: Hmm, removing the BUILD_BUG_ON removes the protection from padding and struct re-ordering of the unnamed struct.  Maybe the better fix is to use a real named struct (maybe kc_shrinker_container) so we can then use container_of in _SHRINKER_FUNCS instead of the raw pointer math.

kmod/src/lock.c
```diff
@@ -1475,7 +1490,10 @@ void scoutfs_free_unused_locks(struct super_block *sb)
 		.nr_to_scan = INT_MAX,
 	};
 
-	linfo->shrinker.shrink(&linfo->shrinker, &sc);
+	// shouldn't need to KC_INIT_SHRINKER_FUNCS as we're calling the shrinker
```
zabbo, line 1493: (Don't use // comments in the kernel.)   And also just remove this comment, it's fine to just make the direct call without commentary.

kmod/src/lock.c
```diff
@@ -1415,24 +1430,25 @@ static void lock_shrink_worker(struct work_struct *work)
  * mode which will prevent the lock from being freed when the null
  * response arrives.
  */
-static int scoutfs_lock_shrink(struct shrinker *shrink,
-			       struct shrink_control *sc)
+static unsigned long lock_scan_objects(struct shrinker *shrink,
```
zabbo, line 1433: This shrinker patch should be merged with the previous one so that we have one shrinker commit where we can review all the changes and see consistency, etc.

kmod/src/kernelcompat.h
```diff
@@ -172,6 +172,18 @@ do {					\
 #define kc_bio_get_errno(bio)			({ (int)((void)(bio), _error_arg); })
 #endif
 
+/*
+ * MS_* flags from <linux/mount.h> should not be used in the kernel
+ * anymore from 4.x onwards. Instead, we need to use the SB_* versions
+ */
+#ifdef SB_POSIXACL
+#define KC_MS_POSIXACL SB_POSIXACL
```
zabbo, line 180: Can we just use the SB_ definitions directly in our code and fall back to the MS_ ones in older code?

kmod/src/Makefile.kernelcompat
```diff
@@ -143,3 +143,22 @@ ifneq (,$(shell grep '(*shrink)' include/linux/shrinker.h))
...
+
+#
+# v4.x
+#
+# linux/iversion.h needs to manually be included for code that
+# manipulates this field.
+#
+ifneq (,$(shell grep 'define _LINUX_IVERSION_H' include/linux/iversion.h))
+ccflags-y += -DKC_NEED_LINUX_IVERSION_H=1
+endif
```
zabbo, line 155: Hmm, this is just a hunk that's out of place?

kmod/src/ioctl.c
```diff
@@ -516,7 +518,9 @@ static long scoutfs_ioc_stage(struct file *file, unsigned long arg)
 	}
 
 	si->staging = true;
+#ifdef KC_LINUX_BACKING_DEV_INFO
 	current->backing_dev_info = mapping->backing_dev_info;
+#endif
```
zabbo, line 525: I don't think this is right.  They removed the bdi from the address space, sure, but we're still required to set the task's bdi when it's calling high level vfs block functions.  Change the = mapping->.. to = inode_to_bdi(inode) and then leave the = NULL.   Take a peek at how/when other file systems also do this upstream.

kmod/src/xattr.c
```diff
@@ -864,18 +865,35 @@ static const char *full_name_hack(void *handler, const char *name, int len)
...
+		 struct inode *inode, const char *name, const void *value,
+		 size_t size, int flags)
+{
+#else
+		(struct dentry *dentry, const char *name,
+		 const void *value, size_t size, int flags, int handler_flags)
 {
 	name = full_name_hack(NULL, name, handler_flags);
+#endif
 	return scoutfs_xattr_set(dentry, name, value, size, flags);
```
zabbo, line 897: Modern code still needs to call scoutfs_xattr_set() with the full xattr name, not just the name after the prefix.  full_name_hack() was doing that.  Modern code has to call xattr_full_name(handler, name).  If we're calling it directly instead of having ifdeffery in full_name_hack then we can drop the NULL handler argument to full_name_hack and update its comment.

kmod/src/Makefile.kernelcompat
```diff
@@ -169,3 +169,10 @@ endif
 ifneq (,$(shell grep 'int...get..const struct xattr_handler.*struct dentry.*dentry,' include/linux/xattr.h))
 ccflags-y += -DKC_XATTR_STRUCT_XATTR_HANDLER=1
 endif
+
+#
+# Network stack changes with v4.x
```
zabbo, line 174: There's too many unrelated interface changes in here.  They should be broken up and triggered by actual interface changes, not presumed association with the presence of the user_msghdr.

kmod/src/kernelcompat.h
```diff
@@ -215,4 +215,16 @@ struct kc_shrinker_funcs {
 } while (0)
 #endif
 
+#ifdef KC_USER_MSGHDR
+#define kc_msghdr user_msghdr
+#define kc_kernel_getsockname(a, b, c) kernel_getsockname(a, b)
```
zabbo, line 238: Like other wrappers, try to use meaningful argument names here so there's just a wee bit more context for our brains when we're bouncing around the code.  Extra credit for adding nop use of args that are dropped so that they don't trigger unused warnings in older code, but the existing codes pretty spotty with that.. best effort, I guess.

zabbo (changes requested):
Lots of great stuff in here.  I didn't comment on commits that either looked fine or had FIXMEs and couldn't possibly work.  A few notes for the commits globally:

Check the commit history,  we stopped using "$WHATEVER:" prefixes in the commits 'cause they don't make much sense in our repo.  So drop all the "compat:" prefixes.  Just use natural brief language to describe the change.

Make sure each compat action has a comment in the code with the git-describe output for the relevant commit and some commentary on what's up.  Follow the existing examples.  Don't use raw commit hashes.

Never use version (sometimes implicitly by associating interfaces) tests.  Distro's whole deal is moving code blocks around regardless of their relation to each other.  Chunks of code are cherry picked in different combinations, leapfrog each other, etc.

I think that's it!

kmod/src/lock.c
```diff
@@ -12,7 +12,9 @@
  */
 #include <linux/kernel.h>
 #include <linux/fs.h>
-#include <linux/preempt_mask.h> /* a rhel shed.h needed preempt_offset? */
+#if defined(RHEL_MAJOR) && RHEL_MAJOR <= 7
```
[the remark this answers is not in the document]
aversecat: Dropping the include and adjusting the commit msg.

kmod/src/kernelcompat.h
```diff
@@ -94,4 +94,10 @@ static inline int dir_emit_dots(struct file *file, void *dirent,
 #define d_materialise_unique(a, b) d_splice_alias(b, a)
 #endif
 
+#ifdef KC_RB_DECLARE_CALLBACKS_MAX
+#define KC_RB_DECLARE_CALLBACKS RB_DECLARE_CALLBACKS_MAX
```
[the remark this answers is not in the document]
aversecat: Even FC37 contains this patch, still:

```
$ grep -r define.RB_DECLARE_CALLBACKS_MAX /usr/src/kernels/
/usr/src/kernels/6.2.9-200.fc37.x86_64/include/linux/rbtree_augmented.h:#define RB_DECLARE_CALLBACKS_MAX(RBSTATIC, RBNAME, RBSTRUCT, RBFIELD,	      \
/usr/src/kernels/4.18.0-425.19.2.el8_7.x86_64/include/linux/rbtree_augmented.h:#define RB_DECLARE_CALLBACKS_MAX(RBSTATIC, RBNAME, RBSTRUCT, RBFIELD,	      \
```

I've found https://www.mail-archive.com/linux-kernel@vger.kernel.org/msg2040737.html as well. I still can't seem to find this patch in the actual rhel repos just yet but it's there. And the latter URL indeed suggests they change all the callers as well. It's indeed bizarre that this is even in 6.2.9, still, while upstream hasn't bothered to adopt this.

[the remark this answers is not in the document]
aversecat: Actually, I need to revisit this.

kmod/src/acl.c
```diff
@@ -12,6 +12,9 @@
  */
 #include <linux/kernel.h>
 #include <linux/fs.h>
+#ifdef KC_NEED_LINUX_IVERSION_H
```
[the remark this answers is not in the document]
aversecat: Trivial, yes, done.

kmod/src/kernelcompat.h
```diff
@@ -100,4 +100,14 @@ static inline int dir_emit_dots(struct file *file, void *dirent,
 #define KC_RB_DECLARE_CALLBACKS RB_DECLARE_CALLBACKS
 #endif
 
+#ifdef KC_INODE_I_RWSEM
+#define kc_inode_lock(inode) down_write(inode->i_rwsem)
```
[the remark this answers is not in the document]
aversecat: After some more reading... `inode_lock()` and friends already existed and used `i_mutex` since like forever, so this becomes much more simple without any `#ifdef` use.

However, there is a name clash in `file.c` where there's a stack variable that needs to be renamed from `inode_lock` so I named it `scoutfs_inode_lock`. A little long but it at least it's simple.

kmod/src/inode.c
```diff
@@ -474,8 +474,11 @@ int scoutfs_setattr(struct dentry *dentry, struct iattr *attr)
 				 SCOUTFS_LKF_REFRESH_INODE, inode, &lock);
 	if (ret)
 		return ret;
-
+#ifndef KC_SETATTR_PREPARE
 	ret = inode_change_ok(inode, attr);
+#else
+	ret = setattr_prepare(dentry, attr);
```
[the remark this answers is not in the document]
aversecat: I don't see this idmap or anything like it anywhere. Is this an inode or dentry thing I just don't see?

I do see v6.x adding a namespace arg to the parameter list of setattr_prepare but that's not a concern for this series yet...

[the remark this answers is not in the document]
aversecat: I did make the other changes, though. Next version will have them.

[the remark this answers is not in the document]
zabbo: > I don't see this idmap or anything like it anywhere. Is this an inode or dentry thing I just don't see?
> 
> I do see v6.x adding a namespace arg to the parameter list of setattr_prepare but that's not a concern for this series yet...

No, you got it, I was seeing it upstream.  Carry on :).

kmod/src/item.c
aversecat: Hmm, so can we get rid of the `min_t(u64, , ULONG_MAX / 2)` here then? Or do we still want to cap at ULONG_MAX / 2 for safety?

zabbo: I think we still want the cap, just so the counters don't wrap, though it does look like a pretty arbitrary and silly cap.

kmod/src/kernelcompat.h
aversecat: :)

The biggest problem I had here is that I just couldn't come up with a way to make this BUILD_BUG_ON() actually compile. Note the typo (`cont`)  but even that isn't the problem. I end up with this wonderful error message salad:

```
rc/block.c:1296:28: error: expected specifier-qualifier-list before ‘block_count_objects’
           &binf->shrinker, block_count_objects, block_scan_objects);
                            ^
include/linux/compiler.h:434:19: note: in definition of macro ‘__compiletime_assert’
   bool __cond = !(condition);    \
                   ^
include/linux/compiler.h:454:2: note: in expansion of macro ‘_compiletime_assert’
  _compiletime_assert(condition, msg, __compiletime_assert_, __LINE__)
  ^
include/linux/bug.h:53:37: note: in expansion of macro ‘compiletime_assert’
 #define BUILD_BUG_ON_MSG(cond, msg) compiletime_assert(!(cond), msg)
                                     ^
include/linux/bug.h:77:2: note: in expansion of macro ‘BUILD_BUG_ON_MSG’
  BUILD_BUG_ON_MSG(condition, "BUILD_BUG_ON failed: " #condition)
  ^
/home/auke/scoutfs/kmod/src/kernelcompat.h:228:2: note: in expansion of macro ‘BUILD_BUG_ON’
  BUILD_BUG_ON(offsetof(countfn, shrinker_funcs) + sizeof(struct kc_shrinker_funcs) != \
  ^
include/linux/stddef.h:16:32: note: in expansion of macro ‘__compiler_offsetof’
 #define offsetof(TYPE, MEMBER) __compiler_offsetof(TYPE, MEMBER)
                                ^
/home/auke/scoutfs/kmod/src/kernelcompat.h:228:15: note: in expansion of macro ‘offsetof’
  BUILD_BUG_ON(offsetof(countfn, shrinker_funcs) + sizeof(struct kc_shrinker_funcs) != \
               ^
/home/auke/scoutfs/kmod/src/block.c:1295:2: note: in expansion of macro ‘KC_INIT_SHRINKER_FUNCS’
  KC_INIT_SHRINKER_FUNCS(struct block_info, shrinker,
  ^
```

I do see what you intend to do here... I'm gonna skip over this for now though.

kmod/src/ioctl.c
aversecat: Yeah! got it.

kmod/src/xattr.c
aversecat: Clearly I didn't even read the comment o_O

kmod/src/kernelcompat.h
zabbo: Yeah, the unnamed struct is gumming things up.  Probably just better to make a _container struct and then use container_of instead of the sketchy manual pointer math..

kmod/src/kernelcompat.h
```diff
@@ -184,6 +184,10 @@ do {					\
 #define KC_MS_I_VERSION MS_I_VERSION
 #endif
 
+#ifndef CURRENT_TIME
+#define CURRENT_TIME (current_kernel_time64())
```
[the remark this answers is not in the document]
aversecat: I think I've got this down, too.

versity-github:
Build 15 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/15/ git: a4ccf2567378eb457776a31344af6bbcdd4bc845

versity-github:
Build 20 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/20/ git: 4b9411680480857ccab85150d1a3ca44d5ddf4ba

versity-github:
Build 25 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/25/ git: 2e57c329bee0fe252cc06d74ecbb2c44d5557617

versity-github:
Build 26 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/26/ git: d7ad6e586648f107eb9908bcce295ca5f34e3212

versity-github:
Build 27 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/27/ git: 5f4aecdd747e77cb12375e9e85db43966968774b

kmod/src/dir.c
```diff
@@ -1138,18 +1139,110 @@ static void scoutfs_put_link(struct dentry *dentry, struct nameidata *nd,
...
+		kfree(path);
+		path = ERR_PTR(ret);
+	}
+	scoutfs_unlock(sb, inode_lock, SCOUTFS_LOCK_READ);
+	return path;
+}
+#endif
+
+// some fwd decls
+static int scoutfs_tmpfile(struct inode *dir, struct dentry *dentry, umode_t mode);
```
zabbo, line 1211: We don't need these fwd decls (and don't need a comment -- they're clear :)).  Just move the struct down under the functions, that's the only reason it is where it is to begin with.

kmod/src/inode.c
```diff
@@ -356,6 +365,25 @@ int scoutfs_getattr(struct vfsmount *mnt, struct dentry *dentry,
 	}
 	return ret;
 }
+#else
+int scoutfs_getattr(const struct path *path, struct kstat *stat,
```
zabbo, line 369: These getattr functions are identical, just ifdef the signatures.  If we ever actually use something with the mask then we can break out the different implementations.

kmod/src/kernelcompat.c
```diff
@@ -21,3 +21,33 @@ int kc_shrink_wrapper(struct shrinker *shrink, struct shrink_control *sc)
...
+
+#ifndef KC_CURRENT_TIME_INODE
+struct timespec64 kc_current_time(struct inode *inode)
+{
+	struct timespec64 now;
+	unsigned gran;
+
+	getnstimeofday64(&now);
+
+	if (unlikely(!inode->i_sb)) {
```
zabbo, line 35: Why does this have a copy of timespec_truncate?  can't we just call it?  It looks like it's consistently exported.

kmod/src/scoutfs_trace.h
```diff
@@ -2020,9 +2020,9 @@ DEFINE_EVENT(scoutfs_quorum_message_class, scoutfs_quorum_recv_message,
 
 TRACE_EVENT(scoutfs_quorum_loop,
 	TP_PROTO(struct super_block *sb, int role, u64 term, int vote_for,
-		 unsigned long vote_bits, struct timespec64 timeout),
+		 unsigned long vote_bits, time64_t timeout_sec, long timeout_nsec),
```
zabbo, line 2023: It's not worth the noise to get the trace to have secs/nsecs.  Just have it take a u64 nsecs and have the trace event caller use ktime_to_ns().

kmod/src/scoutfs_trace.h
```diff
@@ -2031,8 +2031,8 @@ TRACE_EVENT(scoutfs_quorum_loop,
 		__field(int, vote_for)
 		__field(unsigned long, vote_bits)
 		__field(unsigned long, vote_count)
-		__field(unsigned long long, timeout_sec)
```
zabbo, line 2034: The trace events use basic types because of the print format using %llu below.  It'd trigger warnings if the types ever went out of alignment, this can use an unsigned long long nsecs and then print that.

kmod/src/net.c
```diff
@@ -948,22 +948,27 @@ static int sock_opts_and_names(struct scoutfs_net_connection *conn,
...
 		ret = -EAFNOSUPPORT;
-	if (ret)
+	if (ret < 0)
 		goto out;
 
 	addrlen = sizeof(struct sockaddr_in);
-	ret = kernel_getpeername(sock, (struct sockaddr *)&conn->peername,
+	ret = kc_kernel_getpeername(sock, (struct sockaddr *)&conn->peername,
 				 &addrlen);
 	if (ret == 0 && addrlen != sizeof(struct sockaddr_in))
```
zabbo, line 969: This check is now dead.  The hope is always to have the code look like upstream and have compat to backwards in time, not the opposite.  So remove the addrlens from our calls, and have the compat wrapper for the older kernels pass in its own addrlen and return it if the kernel_ call returned 0.  Then these calls will check ret > = 0 && ret != sizeof().

kmod/src/net.c
```diff
@@ -948,22 +948,27 @@ static int sock_opts_and_names(struct scoutfs_net_connection *conn,
...
-	if (ret)
+	if (ret < 0)
 		goto out;
 
 	conn->last_peername = conn->peername;
+	/*
+	 * Because getsock/peername return (>0) as success, we cannot just
+	 * return ret here.
+	 */
+	return 0;
```
zabbo, line 979: The point of the out: nonsense is to have one return statement (that we can then hook into or trace or whatever).  The pattern here is to set ret to 0 before it falls into the label and return, it's all over the place.  Just drop the comment.

kmod/src/ioctl.c
```diff
@@ -516,7 +519,11 @@ static long scoutfs_ioc_stage(struct file *file, unsigned long arg)
 	}
 
 	si->staging = true;
+#ifdef KC_LINUX_BACKING_DEV_INFO
```
zabbo, line 524: Can't we always use inode_to_bdi() and drop all the compat and ifdefs?

kmod/src/kernelcompat.h
zabbo, line 1: There's a comment that uses MS_ as well, need to update it to.

kmod/src/Makefile.kernelcompat
```diff
@@ -122,3 +122,13 @@ endif
...
 endif
+
+#
+# v3.11-8765-ga0b02131c5fc
+#
+# Remove the old ->shrink() API, ->{scan,count}_objects is preferred.
+#
+ifneq (,$(shell grep '(*shrink)' include/linux/shrinker.h))
+ccflags-y += -DKC_SHRINKER_SHRINK
+KC_BUILD_KERNELCOMPAT=1
```
zabbo, line 133: This dead define can be dropped.

kmod/src/block.c
```diff
@@ -1090,24 +1101,20 @@ u64 scoutfs_block_writer_dirty_bytes(struct super_block *sb,
...
 
-	nr = sc->nr_to_scan;
-	if (nr == 0)
-		goto out;
-
-	scoutfs_inc_counter(sb, block_cache_shrink);
+	scoutfs_inc_counter(sb, block_cache_scan_objects);
 
-	nr = DIV_ROUND_UP(nr, SCOUTFS_BLOCK_LG_PAGES_PER);
+	nr = DIV_ROUND_UP(sc->nr_to_scan, SCOUTFS_BLOCK_LG_PAGES_PER);
```
zabbo, line 1117: The changes don't maintain this scaling from object (block) counts to scaled by their number of pages.  It scales to nr on input and then returns freed unscaled.  Let's just remove all this scaling entirely.  All the _count functions can stop doing it, and all the inputs of the scan functions can stop doing it.  I don't think it ever amounted to everything and it's messing us up.

kmod/src/block.c
```diff
@@ -1160,12 +1168,11 @@ static int block_shrink(struct shrinker *shrink, struct shrink_control *sc)
 
 	rhashtable_walk_stop(&iter);
 	rhashtable_walk_exit(&iter);
-out:
+
 	if (stop)
 		return -1;
```
zabbo, line 1173: Returning -1 as an unsigned long should raise eyebrows.  This is the semantics of the old single shrink.  The new scan should return SHRINK_STOP in this case.  The compat wrapper can see that and return -1 for the single int shrink function.

kmod/src/item.c
```diff
@@ -2530,27 +2530,41 @@ void scoutfs_item_invalidate(struct super_block *sb, struct scoutfs_key *start,
...
+{
+	struct item_cache_info *cinf = container_of(shrink,
+						    struct item_cache_info,
+						    shrinker);
+	struct super_block *sb = cinf->sb;
+
+	scoutfs_inc_counter(sb, item_cache_count_objects);
+
+	/* magic numbers as we approach ~0UL :/ */
+	return min_t(u64, (u64)(cinf->lru_pages), ULONG_MAX / 2);
```
zabbo, line 2546: It's annoying that there isn't help for not having the unsigned long count wander up  into the magic SHRINK_{STOP|EMPTY} values.  But here we are.  I wonder if instead of doing this raw min(ulong_max/2) in each scan, if we shouldn't have a static inline helper in util.h that returns the min of the count and LONG_MAX.  Seems cleaner, and a comment around that helper could explain what's going on.

kmod/src/kernelcompat.c
```diff
@@ -0,0 +1,23 @@
...
+ * we have a .shrink() helper that performs the shrink work in terms of
+ * count/scan.
+ */
+int kc_shrink_wrapper(struct shrinker *shrink, struct shrink_control *sc)
+{
+	struct kc_shrinker_funcs *funcs = KC_SHRINKER_FUNCS(shrink);
+	unsigned long nr;
+
+	if (sc->nr_to_scan != 0)
+		funcs->scan_objects(shrink, sc);
```
zabbo, line 19: This should get and check the return of scan and translate SHRINK_STOP into -1 and _EMPTY into 0.  (we don't use empty, but we could, and might as well get it right before we forget that there's a compat thing we have to go check if we update a scan function.)

kmod/src/kernelcompat.h
```diff
@@ -181,4 +181,35 @@ do {					\
...
+	unsigned long (*scan_objects)(struct shrinker *, struct shrink_control *sc);
+};
+/* using adjacent member of an unnamed struct */
+#define KC_DEFINE_SHRINKER(name) struct				\
+	{							\
+		struct kc_shrinker_funcs shrinker_funcs;	\
+		struct shrinker (name);				\
+	}
+#define KC_SHRINKER_FUNCS(shrinkfn) \
+	((void *)((long)(shrinkfn) - sizeof(struct kc_shrinker_funcs)))
```
zabbo, line 224: I think fixing this raw pointer math is pretty easy.  We already have the kc_shrinker_funcs struct.  Rename it to a _shrinker_wrapper and put the struct shrinker in there.  The definition and init macros would change a bit, and the _SHRINKER_FUNCS could turn into a SHRINKER_WRAPPER that uses container_of.   Part of why this looks so weird is that 'shrinkfn' name, it's not a function its the shrinker struct.  Renaming that might make it all a bit clearer.

kmod/src/lock.c
```diff
@@ -1412,24 +1426,24 @@ static void lock_shrink_worker(struct work_struct *work)
...
-	int ret;
 
-	nr = sc->nr_to_scan;
-	if (nr == 0)
-		goto out;
+	scoutfs_inc_counter(sb, lock_scan_objects);
 
 	spin_lock(&linfo->lock);
 
+	nr = DIV_ROUND_UP(sc->nr_to_scan, SCOUTFS_BLOCK_LG_PAGES_PER);
```
zabbo, line 1445: and scaling the number of locks to shrink by the number of pages per block in another cache is not right at all.  (another reason to ditch the scaling.)

kmod/src/Makefile.kernelcompat
zabbo, line 1: The commit summary should be _valid, not _create.  I was super confused reviewing for a while until I figured that out :).

kmod/src/Makefile
zabbo, line 1: This shouldn't be its own commit, it should be merged back with whatever first added kernelcompat.c.  Otherwise it's just not build between that commit and this one.

kmod/src/Makefile.kernelcompat
```diff
@@ -43,3 +43,15 @@ endif
 ifneq (,$(shell grep 'posix_acl_valid.*user_ns,' include/linux/posix_acl.h))
 ccflags-y += -DKC_POSIX_ACL_VALID_USER_NS
 endif
+
+#
+# rhel8+ has a nonstandard RB_DECLARE_CALLBACKS
```
zabbo, line 67: It certainly looks like el8 and upstream both do have this evolution of the rb aug callback generating macros.  Maybe we should chat more about this history here, I wonder if I'm missing something.

kmod/src/tseq.c
```diff
@@ -46,14 +46,18 @@ static struct scoutfs_tseq_entry *tseq_rb_next(struct scoutfs_tseq_entry *ent)
 	return rb_entry(node, struct scoutfs_tseq_entry, node);
 }
 
+#ifdef RB_DECLARE_CALLBACKS_MAX
+static loff_t tseq_compute_total(struct scoutfs_tseq_entry *ent, int __attribute__((__unused__)) flag)
```
zabbo, line 50: This additional arg definitely needs to be used in the newer compute semantics.  It doesn't return the augmented value anymore, it updates the value itself and then can either return to short circuit or not.  Look at the _compute_max template in the rbtree_augmented.h header.  Off the top of my head, I imagine we'd leave this existing  _compute_total function unchanged and then add a _compute_and_set variant that checks the arg (bool exit, not int flags) to either set and return false or skip and return true.  If we have to have a KC macro anyway, we can just pass both functions in and use whichever is needed.

(and fwiw, using `__attribute__((attr))` is almost never the right thing in the kernel 'cause it supports other compilers.. there's usually a __attr equivalent.. __always_unused in this case (you often run into others like __packed, __weak, __deprecated, etc))

zabbo (changes requested):
Looking really strong!   A few things to take care of, but nothing very high risk.

kmod/src/kernelcompat.c
aversecat: This is a port of timespec64_trunc, which truncates a timespec64. The available timespec_trunc() function truncates a struct timespec. Hence the "copy" - the types differ.

kmod/src/ioctl.c
aversecat: Yes and I had missed that RHEL7 even exports it. so, much simpler now.

kmod/src/Makefile
aversecat: It was added well before already. But since it's only useful on rhel8+, nothing would ever fail compiling. Also, it was always mostly empty.

I've moved it to the bottom most patch in the series so that there's no way to cause any issues bisecting.

aversecat: Ohh no I was just iterating over all the commits to make it bisectable on rhel7, and indeed this needs to go into the shrinker count/scan patch, where kernelcompat.c gets created.

kmod/src/tseq.c
aversecat: Okay, I think I have this down now. We are _not_ looking to use a compute callback to calculate the max() of the augmented values or anything, but instead we want to look at and compute the total. So we do not have a use for the RB_DECLARE_CALLBACKS_MAX() macro - it simply defines an extra callback that handles the typical case where we're interested in the max() function of the node's and subtree's augmented value.

Incidentally, every single user in the kernel went to switch to this _MAX() variant afterwards. Which is why you don't find any more uses of the old case where we just need the RB_DECLARE_CALLBACKS() macro - we don't need the extra auto-generated callback function that it generates - we have our own.

In short, we'll continue to use RB_DECLARE_CALLBACKS() and we will now have 2 different tseq_compute_total variants. The old kernels will use the old version which just returns the new augmented computed value for a node (recursing).  The new kernels will use the new variant which updates the total if it changes, and if passed an `exit` flag (bool), will stop propagation by returning a boolean flag.

versity-github:
Build 32 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/32/ git: 36ee4d946bb21605b2d31e7134e5efd42600854f

versity-github:
Build 3 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/3/ git: 36ee4d946bb21605b2d31e7134e5efd42600854f

versity-github:
Build 45 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/45/ git: 941780867a3972ad129e1383223aeb4a692bd461

kmod/src/dir.c
```diff
@@ -1138,19 +1139,73 @@ static void scoutfs_put_link(struct dentry *dentry, struct nameidata *nd,
...
+	} else if (ret == 0 && path[size - 1]) {
+		scoutfs_corruption(sb, SC_SYMLINK_NOT_NULL_TERM,
+				   corrupt_symlink_not_null_term,
+				   "ino %llu last %u",
+				   scoutfs_ino(inode), path[size - 1]);
+		ret = -EIO;
+	}
+
+	if (ret != -EIO)
+		set_delayed_call(done, kfree_link, path);
```
zabbo, line 1198: There's a double free here if ret < 0 && != -EIO.  I don't think we need another copy of what was _follow_link.  Refactor the old _follow_link to remove the nd_set_link and have it just return the allocated pointer of the symlink arget.  Call it, I don't know. get_link_target or whatever.  Then have a small _follow_link method that calls it and calls nd_set_link if it returns a valid pointer.  (like the !IS_ERR_OR_NULL test in put_link).  And then have another small _get_link method that triggers the delayed call with set_delayed_call if it gets a valid pointer that needs to be freed.  Only free in the main get_link_target func if it's returning an error and had it allocated.

kmod/src/file.c
```diff
@@ -111,7 +112,7 @@ ssize_t scoutfs_file_aio_write(struct kiocb *iocb, const struct iovec *iov,
 	if (ret)
 		goto out;
 
-	if (scoutfs_per_task_add_excl(&si->pt_data_lock, &pt_ent, inode_lock)) {
+	if (scoutfs_per_task_add_excl(&si->pt_data_lock, &pt_ent, scoutfs_inode_lock)) {
```
zabbo, line 115: This confused me for a while.  This hunk should be down in the i_mutex->i_rwsem patch where the lock was originally renamed.

kmod/src/util.h
```diff
@@ -17,4 +17,15 @@ static inline void down_write_two(struct rw_semaphore *a,
 	down_write_nested(b, SINGLE_DEPTH_NESTING);
 }
 
+/*
+ * When returning shrinker counts from scan_objects, we should steer
+ * clear of the magic SHRINK_STOP and SHRINK_EMPTY values, which are near
+ * ~0UL values. Hence, we cap count to ~0L, which is arbitarily high
+ * enough to avoid it.
+ */
+static inline unsigned long shrinker_min_t_long(unsigned long count)
```
zabbo, line 26: Making the input long makes sense as the callers are all longs (or ints).  They onl u64s came from scaling the cached block count to pages.  But if we're not doing that anymore then we can use longs everywhere.  If that's the case, then min here doesn't need to be min_t if both count and LONG_MAX are longs.  And all the callers don't need (u64) casts as they call here (which just gets truncated anyway).

zabbo (changes requested):
With these three little things fixed I think we're good!

versity-github:
Build 17 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/17/ git: 2e47106d3943a0419e77347278b54f4fb9fa3176

versity-github:
Build 46 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/46/ git: 2e47106d3943a0419e77347278b54f4fb9fa3176

kmod/src/dir.c
```diff
@@ -1111,46 +1111,55 @@ static void *scoutfs_follow_link(struct dentry *dentry, struct nameidata *nd)
...
 				   corrupt_symlink_not_null_term,
 				   "ino %llu last %u",
 				   scoutfs_ino(inode), path[size - 1]);
+		kfree(path);
 		ret = -EIO;
 	}
 
 out:
-	if (ret < 0) {
-		kfree(path);
```
zabbo, line 1126: This now leaks the allocated path on errors that aren't caught by the two cases that call _corruption().  I'm not sure why the free was removed and put in those two cases, but undoing that change will fix it.

kmod/src/dir.c
```diff
@@ -1111,46 +1111,55 @@ static void *scoutfs_follow_link(struct dentry *dentry, struct nameidata *nd)
...
 
+#ifdef KC_LINUX_HAVE_RHEL_IOPS_WRAPPER
+static void *scoutfs_follow_link(struct dentry *dentry, struct nameidata *nd)
+{
+	char *path;
+
+	path = scoutfs_get_link_target(dentry);
+	if (!IS_ERR_OR_NULL(path))
+		nd_set_link(nd, path);
+	return NULL;
```
zabbo, line 1142: This needs to return path.  Returning NULL masks errors and leaks that path.  It's the return from this which is passed to put_link to be freed.

zabbo (changes requested):
The symlink changes still need work.  With that fixed, and with the pr and periodic tests passing, we should be good.

aversecat:
Pushed fixes for last review.

versity-github:
Build 19 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/19/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

versity-github:
Build 47 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/47/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

zabbo (approved):
Code looks good.  When the tests pass we can land this.

aversecat:
retest

versity-github:
Build 27 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/27/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

versity-github:
Build 54 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/54/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

aversecat:
retest

versity-github:
Build 67 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/67/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

versity-github:
Build 40 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/40/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

aversecat:
retest

aversecat:
retest

versity-github:
Build 41 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/41/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

versity-github:
Build 68 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/68/ git: c0cf9cb840fc2b3a7f1119401402fc1cdfd18bea

aversecat:
rebased.

versity-github:
Build 43 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/43/ git: d2c2fece2ae9ab232c403a8434d9e584aae857b6

versity-github:
Build 70 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/70/ git: d2c2fece2ae9ab232c403a8434d9e584aae857b6

aversecat:
retest

versity-github:
Build 72 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/72/ git: d2c2fece2ae9ab232c403a8434d9e584aae857b6

versity-github:
Build 45 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/45/ git: d2c2fece2ae9ab232c403a8434d9e584aae857b6

versity-github:
Build 73 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/73/ git: d2c2fece2ae9ab232c403a8434d9e584aae857b6

PR #130, opened 2023-07-16
Fix partial preallocation when _contig_only = 0

zabbo:
Data preallocation attempts to allocate large aligned regions of extents.  It tried to fill the hole around a write offset that didn't contain an extent.  It missed the case where there can be multiple extents between the start of the region and the hole. It could try to overwrite these additional existing extents and writes could return EINVAL.

We fix this by trimming the preallocation to start at the write offset if there are any extents in the region before the write offset.

### DIFF
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -458,11 +458,9 @@ static int alloc_block(struct super_block *sb, struct inode *inode,
 		/*
 		 * Preallocation within aligned regions tries to
 		 * allocate an extent to fill the hole in the region
-		 * that contains iblock.  We search for a next extent
-		 * from the start of the region.  If it's at the start
-		 * we might have to search again to find an existing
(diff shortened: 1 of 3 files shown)


versity-github:
Build 29 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/29/ git: 105f1d9e9dafdd323dd57e0b00b07b9dfd21138d

versity-github:
Build 30 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/30/ git: 55f9435fad37b1b34d96914060fd921605303509

zabbo:
(that test failure is a very bizarre error that we've only seen once before, not at all related to these changes.. I'm suspecting infra problems)

aversecat (approved):

PR #139, opened 2023-10-03
Start server commits when holds wait for alloc

zabbo:
Server code that wants to dirty blocks by holding a commit won't be allowed to until the current allocators for the server transaction have enough space for the holder.  As an active holder applies the commit the allocators are refilled and the waiting holders will proceed.

But the current allocators can have no resources as the server starts up.  There will never be active holders to apply the commit and refill the allocators.  In this case all the holders will block indefinitely.

The fix is to trigger a server commit when a holder doesn't have room. It used to be that commits were only triggered when apply callers were waiting.  We transfer some of that logic into a new 'committing' field so that we can have commits in flight without apply callers waiting.  We add it to the server commit tracing.

While we're at it we clean up the logic that tests if a hold can proceed.  It used to be confusingly split across two functions that both could sample the current allocator space remaining.  This could lead to weird cases where the first holder could use the second alloc remaining call, not the one whose values were tested to see if the holder could fit.  Now each hold check only samples the allocators once.

And finally we fix a subtle case where the budget exceeded message can spuriously trigger in the case where dirtying the freed list created a new empty block after the holder recorded the amount of space in the freed block.

### DIFF
--- a/kmod/src/scoutfs_trace.h
+++ b/kmod/src/scoutfs_trace.h
@@ -1896,8 +1896,9 @@ DEFINE_EVENT(scoutfs_server_client_count_class, scoutfs_server_client_down,
 
 DECLARE_EVENT_CLASS(scoutfs_server_commit_users_class,
         TP_PROTO(struct super_block *sb, int holding, int applying, int nr_holders,
-		 u32 avail_before, u32 freed_before, int exceeded),
-        TP_ARGS(sb, holding, applying, nr_holders, avail_before, freed_before, exceeded),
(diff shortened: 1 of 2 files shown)


versity-github:
Build 32 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/32/ git: 4784ccdfd59b57952267a564160aef825b75feed

versity-github:
Build 59 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/59/ git: 4784ccdfd59b57952267a564160aef825b75feed

PR #155, opened 2023-11-13
Fix wrapping detection code blocks.

aversecat:
A new test from xfstests (generic/525) exposes that we poorly detect wrapping of u64/s64 values in a few cases. The key test is that we should always be able to fully use the last block, where (offset + length) is larger than (e.g.) LLONG_MAX. In other words, we want to make sure length is not negative, and (offset + (length - 1)) doesn't wrap.

scoutfs_client_alloc_inodes() checks that u64 inode numbers do not wrap. We allocate these in 10*1024 hunks, and in practice we will never hit the issue since (2^64 % 10240) is not 0.

scoutfs_fallocate() may hit the wraping issue.

scoutfs_data_wait_check() is the case that was exposed by generic/525.

Our own test `simple-release-extents` almost exposed the wrapping issue in scoutfs_ioc_release(). We adjust the test to test the edge case and the wrapping case properly.

### DIFF
--- a/kmod/src/client.c
+++ b/kmod/src/client.c
@@ -80,7 +80,7 @@ int scoutfs_client_alloc_inodes(struct super_block *sb, u64 count,
 
 		if (*nr == 0)
 			ret = -ENOSPC;
-		else if (*ino + *nr < *ino)
+		else if ((*ino + (*nr - 1) < *ino) || (*nr < 0))
 			ret = -EINVAL;
 	}
 
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -1071,7 +1071,7 @@ long scoutfs_fallocate(struct file *file, int mode, loff_t offset, loff_t len)
 	}
 
 	/* catch wrapping */
-	if (offset + len < offset) {
+	if ((len < 0) || (offset + (len - 1) < offset)) {
 		ret = -EINVAL;
 		goto out;
 	}
@@ -1723,7 +1723,8 @@ int scoutfs_data_wait_check(struct inode *inode, loff_t pos, loff_t len,
 	if (WARN_ON_ONCE(sef & SEF_UNKNOWN) ||
 	    WARN_ON_ONCE(op & SCOUTFS_IOC_DWO_UNKNOWN) ||
 	    WARN_ON_ONCE(dw && !RB_EMPTY_NODE(&dw->node)) ||
-	    WARN_ON_ONCE(pos + len < pos)) {
+	    WARN_ON_ONCE(len < 0) ||
+	    WARN_ON_ONCE(pos + (len - 1) < pos)) {
 		ret = -EINVAL;
 		goto out;
 	}
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -1717,6 +1717,9 @@ int scoutfs_data_wait_check(struct inode *inode, loff_t pos, loff_t len,
 	u64 off;
 	int ret = 0;
 
+	if (len == 0)
+		goto out;
+
 	if (WARN_ON_ONCE(sef & SEF_UNKNOWN) ||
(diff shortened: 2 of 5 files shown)


versity-github:
Build 78 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/78/ git: 9966028fd621d078473d29b812a6400f4210e88e

versity-github:
Build 103 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/103/ git: 9966028fd621d078473d29b812a6400f4210e88e

versity-github:
Build 79 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/79/ git: c6b4d03096d826a871f18c79533a237a5b8f8ec8

versity-github:
Build 104 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/104/ git: c6b4d03096d826a871f18c79533a237a5b8f8ec8

aversecat:
Swapped order to maintain bisectability. Also rebased.

versity-github:
Build 81 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/81/ git: 6da2188ac8c61b148ee1e2c92fca2cd4a4a571cd

versity-github:
Build 106 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/106/ git: 6da2188ac8c61b148ee1e2c92fca2cd4a4a571cd

zabbo (changes requested):
I think the root cause of these mistakes is that I was open coding these tests rather than using a semantic function that might have made me stop and think if it was the right thing.  (The < a construct I think came from unrelated but kind of similar tests for if changing precision drops data ((type)a < a).   Anyway, let's not continue the mistake by continuing to open code these tests.  Then we need to be careful not to use the overflowed result to discover the overflow because compilers can remove them :(.  There's also two more cases I saw in ioctl.c in scoutfs_ioc_move_blocks().  (and feel free to remove comments about checking for wrapping once the function makes it clear that's what we're doing.)  This seems like it'd work to me, can you work it in and see if it makes sense and works for you?

```
/*
 * Returns true if the contents of the region described by offset and
 * length wrap at the precision of the offset's type.
 *
 * The next offset after the region can wrap to 0 and this will return
 * false because the region did not wrap.  (1,~0 does not wrap, 2,~0
 * does).
 *
 * This is fiddly because signed int overflow is undefined behavior in C
 * and compilers have optimized away tests that rely on the result of
 * undefined addition that overflows.
 *
 * We need to subtract 1 from the len for the overflow test to tell us
 * that the region wrapped and not the next offset after the region.  We
 * take liberties and say that regions of len == 0 do not wrap and len <
 * 0 always wrap.  This lets us ensure that the len is positive which
 * devolves the signed overflow test into the unsigned overflow test,
 * just with a signed max.
 */
#define off_len_region_wraps(off, len)                                  \
({                                                                      \
        __typeof__(off) _off = (off);                                   \
        __typeof__(off) _len = (len);                                   \
        __typeof__(off) _max = ~0ULL;                                   \
        bool _wrap;                                                     \
                                                                        \
        BUILD_BUG_ON(sizeof(__typeof__(len)) > sizeof(__typeof__(off)));\
                                                                        \
        if (_len == 0) {                                                \
                _wrap = false;                                          \
        } else if (_len < 0) {                                          \
                _wrap = true;                                           \
        } else  {                                                       \
                 if ((__typeof__(off))(~0ULL) < 0)                      \
                        _max = (unsigned)_max >> 1;                     \
                _wrap = (_max - _off) < (_len - 1);                     \
        }                                                               \
                                                                        \
        _wrap;                                                          \
})
```

zabbo:
Hmm, no, that (unsigned) isn't quite right to drop the signed bit.   Hmm.

aversecat:
Closing this one here. I'll have something in the `el9` patch set instead (where it's exposed by an xfstests test).

PR #168, opened 2024-03-20
Scoutfs Check

h-shaffer:
Added checks for superblock fields, quorum config, and quorum slot addressing


utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+		debug("Quorum slot %u family: %u\n", i, family);
+		if (family != SCOUTFS_AF_IPV4)
+			problem(PB_QSLOT_BAD_FAM, "Port %u has unsupported family %u", port, family);
+
+		in.s_addr = htonl(le32_to_cpu(slots[i].addr.v4.addr));
+		debug("Quorum Slot %u ipv4 address: %s\n", i, inet_ntoa(in));
+		if ((in.s_addr & 0xff) != 0xff)
+			problem(PB_QSLOT_BAD_ADDR, "Quorum Slot %u has multicast address", i);
+		else if ((in.s_addr & 0xff000000) != 0)
+			problem(PB_QSLOT_NO_ADDR, "Quorum Slot %u has reserved address, cannot start with 0", i);
```
aversecat, line 97: s/NO/BAD/ here?

utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+		family = le16_to_cpu(slots[i].addr.v4.family);
+		debug("Quorum slot %u family: %u\n", i, family);
+		if (family != SCOUTFS_AF_IPV4)
+			problem(PB_QSLOT_BAD_FAM, "Port %u has unsupported family %u", port, family);
+
+		in.s_addr = htonl(le32_to_cpu(slots[i].addr.v4.addr));
+		debug("Quorum Slot %u ipv4 address: %s\n", i, inet_ntoa(in));
+		if ((in.s_addr & 0xff) != 0xff)
+			problem(PB_QSLOT_BAD_ADDR, "Quorum Slot %u has multicast address", i);
+		else if ((in.s_addr & 0xff000000) != 0)
```
aversecat, line 96: This looks wrong. The high order byte may not be `0`. In other words, if the byte equals zero, there is a problem. You have `!=` here instead.

utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+			problem(PB_QSLOT_BAD_PORT, "Quorum Slot %u has bad port", i);
+
+		family = le16_to_cpu(slots[i].addr.v4.family);
+		debug("Quorum slot %u family: %u\n", i, family);
+		if (family != SCOUTFS_AF_IPV4)
+			problem(PB_QSLOT_BAD_FAM, "Port %u has unsupported family %u", port, family);
+
+		in.s_addr = htonl(le32_to_cpu(slots[i].addr.v4.addr));
+		debug("Quorum Slot %u ipv4 address: %s\n", i, inet_ntoa(in));
+		if ((in.s_addr & 0xff) != 0xff)
```
aversecat, line 94: Same problem as below. If the low order byte equals 0xff, there is a problem. Now it raises a problem for 4.4.4.4, which is a valid address.

utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+		debug("Quorum slot %u port: %d\n", i, port);
+		if ((port <= 0) || (port > USHRT_MAX))
+			problem(PB_QSLOT_BAD_PORT, "Quorum Slot %u has bad port", i);
+
+		family = le16_to_cpu(slots[i].addr.v4.family);
+		debug("Quorum slot %u family: %u\n", i, family);
+		if (family != SCOUTFS_AF_IPV4)
+			problem(PB_QSLOT_BAD_FAM, "Port %u has unsupported family %u", port, family);
+
+		in.s_addr = htonl(le32_to_cpu(slots[i].addr.v4.addr));
```
aversecat, line 92: do you really need `ntohl` here?

aversecat (changes requested):

utils/src/check/super.c
h-shaffer: I thought that in.s_addr had to be in network byte order to be properly parsed by intet_ntoa()?

utils/src/check/super.c
h-shaffer: ??

utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+			problem(PB_QSLOT_BAD_ADDR, "Quorum Slot %u has multicast address", i);
+		else if ((in.s_addr & 0xff000000) != 0)
+			problem(PB_QSLOT_NO_ADDR, "Quorum Slot %u has reserved address, cannot start with 0", i);
+		else if (!in.s_addr)
+			problem(PB_QSLOT_NO_ADDR, "Quorum Slot %u has not been assigned ipv4 address", i);
+	}
+	return 0;
+}
+
+static int check_quorum_config(struct scoutfs_quorum_config *qconf)
```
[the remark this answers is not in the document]
h-shaffer: 👍

utils/src/check/super.c
```diff
@@ -56,6 +56,15 @@ int check_super_crc(bool repair)
 	return 0;
 }
 
+static int check_super_flag(__le64 flags)
+{
+	debug("Superblock flag: %llu\n", le64_to_cpu(flags));
+	if (!(flags & cpu_to_le64(SCOUTFS_FLAG_IS_META_BDEV) && (flags &~ 0)))
```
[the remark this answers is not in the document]
h-shaffer: Am I still able to do bitwise operations on a cpu value?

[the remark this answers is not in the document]
aversecat: yes, always. The compiler will error if it can't.

utils/src/check/super.c
aversecat: If the leftmost byte is zero, the IP address looks like `0.x.x.x`. That's a bad address since it's reserved.

Because you haven't checked (here) yet that x.x.x is or is not 0.0.0, this is a little confusing. In essence, you should check for 0.0.0.0 first. Then after that, check for 0.x.x.x and check for x.x.x.255. So changing the order here is also needed.

utils/src/check/super.c
aversecat: well one way to figure this out is to run the code and test it (and maybe test-print the values) to see if it's correct.

utils/src/check/super.c
```diff
@@ -65,15 +68,53 @@ static int check_super_flag(__le64 flags)
...
+
+	for (int i = 0; i < SCOUTFS_QUORUM_MAX_SLOTS; i++) {
+		port = le16_to_cpu(slots[i].addr.v4.port);
+		debug("Quorum slot %u port: %d\n", i, port);
+		if ((port <= 0) || (port > USHRT_MAX))
+			problem(PB_QSLOT_BAD_PORT, "Quorum Slot %u has bad port", i);
+
+		family = le16_to_cpu(slots[i].addr.v4.family);
+		debug("Quorum slot %u family: %u\n", i, family);
+		if (family != SCOUTFS_AF_IPV4)
```
aversecat, line 89: Take a look at `util/quorum.c` at `bool valid_quorum_slots` and you can see how the code uses the `family` bit to determine whether a quorum slot is used or not using `SCOUTFS_AF_NONE`. We need to do the same here.

aversecat (changes requested):

utils/src/check/super.c
```diff
@@ -65,6 +65,18 @@ static int check_super_flag(__le64 flags)
 	return 0;
 }
 
+static int check_quorum_config(struct scoutfs_quorum_config qconf)
```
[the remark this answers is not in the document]
h-shaffer: added this in the wrong patch, will move it back here

aversecat:
I think at this point the code is slowly looking better but all in all it's a lot of functions that are just minuscule. Perhaps we're better off making a simple chain of checks and putting it all in a single function, as it's not really that much.

I'm thinking that ballooning our functions to this much detail isn't going to help at all. It's just too much lines wasted on function calls while the actual checks are 1-2 lines.

aversecat:
I have updated the base branch to include a ton of new things, so you will need to rebase.

aversecat:
One more time. And I have to pull in zach's changes as well.

aversecat:
Maybe just wait to rebase until I have those merged.

aversecat:
Not sure how my own commits ended up in your branch. You may need to rebase again to fix this.

The other thing I forgot to mention is that one of the reasons I want you to use `global_super` instead of `super` is because if we are going to repair things, we should repair them in `global_super` since we have a way to commit changes in it back to disk in a sane way - the cache object `super` is just ephemereal - changes in it are lost.

aversecat:
Can you go through each comment and "mark as resolved" the ones that you have modified to code. We are getting close to wrapping this up and before that we should really make sure we've resolved every issue. You're the one that decides that you have resolved everything.

When we get to a point that I have no more comments left, it's good to merge.

aversecat (approved):

PR #169, opened 2024-04-01
Hunter/check

h-shaffer:
Adding clobbers and repairs for superblock fields


utils/src/check/super.c
```diff
@@ -241,8 +241,18 @@ int check_supers(int data_fd)
 	debug("super magic 0x%04x", global_super->hdr.magic);
 
 	debug("Superblock flag: %llu", global_super->flags);
-	if (global_super->flags != SCOUTFS_FLAG_IS_META_BDEV)
+	if (global_super->flags != SCOUTFS_FLAG_IS_META_BDEV) {
 		problem(PB_SB_BAD_FLAG, "Bad flag: %llu expecting: 1 or 0", global_super->flags);
+		if (repair) {
+			global_super->flags &= SCOUTFS_FLAG_IS_META_BDEV;
```
aversecat, line 247: `global_super` is always set for the meta device. So I would just assign the flag, instead of modifying the incorrect value. The repaired value must always be equal to `SCOUTFS_FLAG_IS_META_BDEV`.

utils/src/check/super.c
```diff
@@ -241,8 +241,18 @@ int check_supers(int data_fd)
...
+			ret =super_commit();
+			if (ret < 0) {
+				fprintf(stderr, "error writing superblock\n");
+				goto out;
+			} else
+				debug("Superblock flag repaired: %llu", global_super->flags);
```
aversecat, line 253: You should call `correct(PB_SB_BAD_FLAG);` here to increment the counter for the corrected problem.

utils/src/check/super.c
```diff
@@ -241,8 +241,18 @@ int check_supers(int data_fd)
 	debug("super magic 0x%04x", global_super->hdr.magic);
 
 	debug("Superblock flag: %llu", global_super->flags);
-	if (global_super->flags != SCOUTFS_FLAG_IS_META_BDEV)
+	if (global_super->flags != SCOUTFS_FLAG_IS_META_BDEV) {
 		problem(PB_SB_BAD_FLAG, "Bad flag: %llu expecting: 1 or 0", global_super->flags);
+		if (repair) {
+			global_super->flags &= SCOUTFS_FLAG_IS_META_BDEV;
+			ret =super_commit();
```
aversecat, line 248: space

utils/src/check/clobber-functions.c
```diff
@@ -111,12 +111,29 @@ static struct clobber_function clobber_pb_sb_hdr_crc_invalid = {
...
+ */
+static int do_clobber_pb_sb_bad_flag(char *data)
+{
+	global_super->flags |= 0xfc;
+	return super_commit();
+}
+
+static struct clobber_function clobber_pb_sb_bad_flag = {
+	PB_SB_BAD_FLAG,
+	"Sets an invalid flag in the superblock.\n" \
```
aversecat, line 125: There are many fields in the superblock, so this is a little ambiguous in the way it's written. More precise would be:

"Sets an invalid value for the flags field in the superblock."

utils/src/check/super.h
```diff
@@ -4,7 +4,7 @@
 extern struct scoutfs_super_block *global_super;
 
 int check_super_crc(bool repair);
-int check_supers(int data_fd);
+int check_supers(int data_fd, bool repair);
```
aversecat, line 7: If you are adding a clobber that also can be repaired, you should modify `tests/test/clobber-repair` and include the clobber in the array `CLOBBERS` so that automated testing can be performed on the clobber and repair code.

utils/src/check/clobber-functions.c
```diff
@@ -127,6 +127,19 @@ static struct clobber_function clobber_pb_sb_bad_flag = {
 	&do_clobber_pb_sb_bad_flag,
 };
 
+static int do_clobber_pb_qconf_wrong_version(char *data)
+{
+	global_super->qconf.version &= 0;
```
aversecat, line 132: `&= 0` is the same as `= 0`. Don't make things needlessly complicated here. Just assign `0` to the field directly.

utils/src/check/super.c
```diff
@@ -262,8 +262,19 @@ int check_supers(int data_fd, bool repair)
 			SCOUTFS_FORMAT_VERSION_MAX);
 
 	debug("Quorum Config Version: %llu", global_super->qconf.version);
-	if (global_super->qconf.version != 1)
+	if (global_super->qconf.version != 1) {
 		problem(PB_QCONF_WRONG_VERSION, "Wrong Version: %llu (expected 1)", global_super->qconf.version);
+		if(repair) {
```
aversecat, line 267: space

utils/src/check/super.c
```diff
@@ -262,8 +262,19 @@ int check_supers(int data_fd, bool repair)
...
+			global_super->qconf.version = 1;
+			ret = super_commit();
+			if (ret < 0) {
+				fprintf(stderr, "error writing superblock\n");
+				goto out;
+			} else
+				debug("Qconf Version repaired: %llu", global_super->qconf.version);
+		}
+
```
aversecat, line 276: unneeded extra line

utils/src/check/super.c
```diff
@@ -262,8 +262,19 @@ int check_supers(int data_fd, bool repair)
...
+	if (global_super->qconf.version != 1) {
 		problem(PB_QCONF_WRONG_VERSION, "Wrong Version: %llu (expected 1)", global_super->qconf.version);
+		if(repair) {
+			global_super->qconf.version = 1;
+			ret = super_commit();
+			if (ret < 0) {
+				fprintf(stderr, "error writing superblock\n");
+				goto out;
+			} else
+				debug("Qconf Version repaired: %llu", global_super->qconf.version);
```
aversecat, line 274: Same comment here as in the other patch: `correct()` the problem here as well.

Plus please add the clobber to the test script.

utils/src/check/problem.c
```diff
@@ -21,6 +21,7 @@ char *prob_strs[] = {
 	PROB_STR(PB_QSLOT_NO_ADDR),
 	PROB_STR(PB_QSLOT_BAD_ADDR),
 	PROB_STR(PB_DATA_DEV_SB_INVALID),
+	PROB_STR(PB_QCONF_WRONG_VERSION),
```
aversecat, line 24: it was already in the list. at line `18`

aversecat (changes requested):

utils/src/check/super.c
aversecat: There's still a missing space here unless github is showing me something outdated...

utils/src/check/super.c
aversecat: Well yes, but, `correct()` already calls `debug()` in its own codepath. So you can remove the debug print out.

utils/src/check/super.c
aversecat: It's still missing.

aversecat (changes requested):

utils/src/check/super.c
h-shaffer: Yea I think that might be it, hopefully my next push will fix it

h-shaffer: Okay I think I fixed it, it seems I fixed it in a later branch but never applied it here, should be fixed her in a second

aversecat (changes requested):

utils/src/check/clobber-functions.c
```diff
@@ -140,6 +140,145 @@ static struct clobber_function clobber_pb_qconf_wrong_version = {
...
+
+	/* pick random non empty qslot to clobber */
+	do {
+		rand_slot = random() & (SCOUTFS_QUORUM_MAX_SLOTS - 1);
+		if (slots[rand_slot].addr.v4.family != SCOUTFS_AF_NONE)
+			slot_empty = false;
+	} while (slot_empty);
+	slots[rand_slot].addr.v4.port = 0;
+	return super_commit();
+}
```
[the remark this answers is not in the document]
h-shaffer: I was thinking of something along the same lines, as all of these functions are near identical

utils/src/check/clobber-functions.c
```diff
@@ -111,12 +111,187 @@ static struct clobber_function clobber_pb_sb_hdr_crc_invalid = {
...
+	srandom(time(NULL));
+
+	/* pick random non empty qslot to clobber */
+	do {
+		rand_slot = random() & (SCOUTFS_QUORUM_MAX_SLOTS - 1);
+		printf("%u\n", rand_slot);
+		if (slots[rand_slot].addr.v4.family != SCOUTFS_AF_NONE)
+			slot_empty = false;
+	} while (slot_empty);
+	slots[rand_slot].addr.v4.family = SCOUTFS_AF_IPV6;
```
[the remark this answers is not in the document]
h-shaffer: I agree, however I was also thinking of collapsing them all down into one function which would eliminate duplication

utils/src/check/clobber-functions.c
```diff
@@ -140,6 +140,53 @@ static struct clobber_function clobber_pb_qconf_wrong_version = {
...
+
+	for (int i = 0; i < SCOUTFS_QUORUM_MAX_SLOTS; i++) {
+		family = le16_to_cpu(global_super->qconf.slots[i].addr.v4.family);
+		if (family == SCOUTFS_AF_NONE)
+			continue;
+
+		filled_slots[j] = i;
+		j++;
+	}
+	return filled_slots[(random() & (j-1))];
```
aversecat, line 158: okay, but, use `%` instead here. While the `&` will work it doesn't create a truly random contstraint and will quite likely never pick some of the slots. For instance, if there are 5 quorum slots, the result is `random() & 4` which isn't ever going to result in `3`. The `%` operator however will do what you want.

utils/src/check/clobber-functions.c
```diff
@@ -140,6 +140,53 @@ static struct clobber_function clobber_pb_qconf_wrong_version = {
...
+}
+
+/*
+ * Clobber a random quorum slot address by setting it to 0
+ */
+static int do_clobber_pb_qslot_bad_addr(char *data)
+{
+	struct scoutfs_quorum_slot *slots = global_super->qconf.slots;
+	int rand_slot = pick_random_qslot();
```
aversecat, line 167: This is nice and clean already. But, you could go even further and make the function just return a pointer to the slot directly:
```
struct scoutfs_quorum_slot *slot = pick_random_qslot();
```

The pick_random_slot() then just needs changing so that it returns `global_super->qconf.slots[<random pick>]` or something like that.

tests/tests/clobber-repair.sh
```diff
@@ -14,6 +14,7 @@ CLOBBERS=(
 "PB_SB_HDR_CRC_INVALID,"
 "PB_SB_BAD_FLAG,"
 "PB_QCONF_WRONG_VERSION"
+"PB_QSLOT_BAD_ADDR -D family",
```
aversecat, line 17: This is incorrectly formatted. Maybe the description above this list isn't verbose enough, but you want this line to look exactly like this:
```
"PB_QSLOT_BAD_ADDR,family"
```

Notice the location of the `,` - it's *within* the quotes.

utils/src/check/clobber-functions.c
```diff
@@ -140,6 +140,57 @@ static struct clobber_function clobber_pb_qconf_wrong_version = {
...
+ */
+static struct scoutfs_quorum_slot *pick_random_qslot(void)
+{
+	int j = 0;
+	int filled_slots[SCOUTFS_QUORUM_MAX_SLOTS];
+	int rand_slot;
+	uint16_t family;
+	srandom(time(NULL));
+
+	for (int i = 0; i < SCOUTFS_QUORUM_MAX_SLOTS; i++) {
```
aversecat, line 154: This won't compile on `el7` because `for (int i` isn't allowed just yet there. You'll have to declare `int i;` above at the top of the function.

utils/src/check/clobber-functions.c
```diff
@@ -140,6 +140,57 @@ static struct clobber_function clobber_pb_qconf_wrong_version = {
...
+		family = le16_to_cpu(global_super->qconf.slots[i].addr.v4.family);
+		if (family == SCOUTFS_AF_NONE)
+			continue;
+
+		filled_slots[j] = i;
+		j++;
+	}
+	rand_slot = filled_slots[(random() % (j-1))];
+	return &global_super->qconf.slots[rand_slot];
```
aversecat, line 163: This doesn't actually work and causes a `Program received signal SIGFPE, Arithmetic exception.` because of the bug here.

First, you're correctly counting valid/used slots with `j` here,but then you use `% (j - 1)` but the modulo operator doesn't work like that. You need to `% j`. Then there's the bug that *none* of the quorum slots could actually be in use, and so your code could run and find `0` filled slots, and then do `rand_slot = filled_slots[random() % -1]` which I'm pretty sure is wildly out of bounds.

Basically, `X & Y` results in a value `between 0 and Y - 1`. If you have an array `[0,j - 1]`, then `random() % j` does the trick.

But you still have to validate that `j` is actually not `0` and within bounds.

tests/tests/clobber-repair.sh
aversecat: Once you add or modify the tests, the output of the test script will no longer match the `golden` output value. If you add a new clobber here in the array, you also have to modify `tests/golden/clobber-repair` to contain the new expected output so that it matches.

aversecat (changes requested):
Try and make sure from now on that your code actually compiles cleanly and runs correctly, and passes tests - before you push.

utils/src/check/clobber-functions.c
h-shaffer: okay, I did this for indexing the slots in super.c as well, should I change it there to?

aversecat: Yes.

aversecat:
Merged into #170

PR #172, opened 2024-04-23
Allocate enough space on disk for log_merge

h-shaffer:
Make sure that we have enough free blocks for the log trees to
    be merged once we finish mounting the system

### DIFF
--- a/utils/src/parallel_restore.c
+++ b/utils/src/parallel_restore.c
@@ -522,6 +522,7 @@ static spr_err_t insert_extent_item(struct scoutfs_parallel_restore_writer *wri,
 
 	err = bti_alloc(sizeof(struct scoutfs_data_extent_val), &bti);
 	if (!err) {
+		bti->key = key;
 		dv = bti->val;
 		dv->blkno = 0;
 		dv->flags = SEF_OFFLINE;
--- a/utils/src/parallel_restore.c
+++ b/utils/src/parallel_restore.c
@@ -1,0 +1,1871 @@
+#include <unistd.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/time.h>
+#include <uuid/uuid.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <assert.h>
+#include <math.h>
+#include <sys/uio.h>
+
+#include "sparse.h"
+#include "util.h"
+#include "format.h"
+#include "crc.h"
+#include "rand.h"
+#include "key.h"
+#include "bitops.h"
+#include "btree.h"
+#include "leaf_item_hash.h"
+#include "name_hash.h"
+#include "mode_types.h"
+#include "srch.h"
+#include "bloom.h"
+
+#include "parallel_restore.h"
+
+#include "list.h"
+#include "lk_rbtree_wrapper.h"
+
+/*
+ * XXX
+ *  - interface versioning?
+ *  - next seq and next ino are both max ino + 1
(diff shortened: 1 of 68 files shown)


utils/src/parallel_restore.c
```diff
@@ -1136,6 +1136,7 @@ static spr_err_t prepare_alloc_builders(struct scoutfs_parallel_restore_writer *
 	dprintf("starting prepare with start %llu len %llu\n", wri->meta_start, wri->meta_len);
 
 	skip = ALLOC_BLOCKS + (SCOUTFS_ALLOC_LIST_MAX_BLOCKS * 2);
+	skip += le64_to_cpu(wri->super.total_meta_blocks) / 10;
```
zabbo, line 1139: "skip" here is the number of blocks that will be used to store allocation structures.  It reflects the number of blocks written by this routine.  By adding the reservation to skip here you're saying that it won't be available as a free record somewhere (either in list block entries or extent items).  Look at the very next thing done with the skip value, it's subtracted from a free extent that's recorded.   So we want to add it to a free extent somewhere.  The next problem is that you can't just add to any old length.  You need to think about the specific extent value that you're removing with the initial / 10 reservation.  It's the end of the block space.  To return that reservation you have to specifically extend the final extent or create a new one.

utils/src/parallel_restore.c
```diff
@@ -1621,6 +1622,15 @@ spr_err_t scoutfs_parallel_restore_init_slices(struct scoutfs_parallel_restore_w
 					       int nr)
 {
 	u64 total = le64_to_cpu(wri->super.total_meta_blocks);
+	/*
+	 * We have to reserve some space (10%) so there are enough free blocks after
+	 * mounting for log_merge to complete
+	 * XXX could possibly take in an arg determining percentage of space to reserve
```
zabbo, line 1628: Can just drop the XXX comment here, there's no reason to call out the universal truth that we could always extend behaviour.  I'd also move the rest of the comment up to a define that contains the reservation percentage which is then used in place of the two raw instances of 10.  And 10 is probably too much, maybe 5?

zabbo (changes requested):

utils/src/parallel_restore.c
h-shaffer: What would the benefit be for adding a new extent onto the end vs extending the final extent?

zabbo: If you extend an existing extent you don't have to create a new one, but you have to identify the extent to extend.

h-shaffer: Could you elaborate a bit more on identifying an extent, I see that there is some math identifying its bounds, start and len, would I just be able to calculate the end of the reserved space and add onto it?

h-shaffer: Could you elaborate a bit more on identifying an extent, I see there's some math there identifying its bounds, couldn't I just calculate the end of the free extent and add a percentage of blocks into the extent?

utils/src/parallel_restore.c
```diff
@@ -1146,6 +1154,13 @@ static spr_err_t prepare_alloc_builders(struct scoutfs_parallel_restore_writer *
 	if (err)
 		goto out;
 	wri->meta_len -= len;
+	/* store reserved extra space a free extent*/
+	end = start + len;
+	extra = le64_to_cpu(wri->super.total_meta_blocks) / EXTRA_BLOCK_SLICE;
+	err = insert_free_items(&wri->meta_btb[0], end, extra);
```
aversecat, line 1160: could just write `start + len` inside the function call here and avoid needing to declare `end` entirely and pre-calculate.

aversecat:
I've merged the two relevant changes into my tree.

PR #174, opened 2024-06-10
Add tracepoint as block read returns ESTALE

zabbo:
Block reads can return ESTALE naturally as mounts read through old cached blocks.  We won't always log it as an error but we should add a tracepoint that can be inspected.


versity-github:
Build 168 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/168/ git: 48716461e4cc78b239546d9047635d11d3ddf13f

aversecat (approved):

PR #179, opened 2024-07-11
Extra device checks when changing formatversion.

aversecat:
This is overly cautious but should help to avoid most common user errors when passing meta and data device nodes.

We check for dupes, and meta flag set/unset for meta/data device respectively.

### DIFF
--- a/utils/src/change_format_version.c
+++ b/utils/src/change_format_version.c
@@ -94,6 +94,18 @@ static int do_change_fmt_vers(struct change_fmt_vers_args *args)
 		goto out;
 	}
 
+	if ((le64_to_cpu(meta_super->flags) & SCOUTFS_FLAG_IS_META_BDEV) == 0) {
+		printf("device argument #1 is not a meta device (swap arguments?)\n");
+		ret = -EINVAL;
+		goto out;
+	}
+
+	if ((le64_to_cpu(data_super->flags) & SCOUTFS_FLAG_IS_META_BDEV) != 0) {
+		printf("device argument #2 is not a data device (swap arguments?)\n");
+		ret = -EINVAL;
+		goto out;
+	}
+
 	if (le64_to_cpu(meta_super->fmt_vers) < SCOUTFS_FORMAT_VERSION_MIN ||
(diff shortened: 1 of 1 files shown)


versity-github:
Build 179 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/179/ git: d53118df500354a4d223333782eff7ca4c900372

versity-github:
Build 154 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/154/ git: d53118df500354a4d223333782eff7ca4c900372

utils/src/change_format_version.c
```diff
@@ -51,6 +51,12 @@ static int do_change_fmt_vers(struct change_fmt_vers_args *args)
 	int data_fd = -1;
 	int ret;
 
+	if (strcmp(args->meta_device, args->data_device) == 0) {
```
zabbo, line 54: Just drop this.  If we had a real problem with having duplicate devices then we'd want something a lot more than string comparison.  It gets very difficult because they could have messed up all sorts of device configuration between path strings and block devices.  But this isn't an important problem -- if you provide the same device then it will always fail the meta/data flag checks.  That's enough to trigger people to stare at their args (/environment) to try and figure out why the devices aren't lining up.

zabbo (changes requested):

utils/src/change_format_version.c
aversecat: Yeah, agreed this isn't needed.

versity-github:
Build 181 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/181/ git: 267c1cc2d5ba2844eef9704209be6ff9790f1541

versity-github:
Build 157 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/157/ git: 267c1cc2d5ba2844eef9704209be6ff9790f1541

zabbo:
The -el8 failure is unrelated, probably a test race, so I'm merging this.

zabbo (approved):

PR #183, opened 2024-09-10
Don't pass data version to attr_x unless the ioctl means to set it.

aversecat:
A small bug in the translation code from setattr_more to attr_x caused setattr to fail if a size zero was passed but no data version was passed.

Add a small test case for this. Without the fix this will produce the following error output ,and nothing with the fix:

  == i_size zero should not require data version to be set
  setattr_more ioctl failed on '/mnt/test.0/test/setattr_more/file': Invalid argument (22)
  scoutfs: setattr failed: Invalid argument (22)

### DIFF
--- a/kmod/src/ioctl.c
+++ b/kmod/src/ioctl.c
@@ -674,8 +674,8 @@ static long scoutfs_ioc_setattr_more(struct file *file, unsigned long arg)
 		goto out;
 	}
 
-	iax->x_mask = SCOUTFS_IOC_IAX_DATA_VERSION | SCOUTFS_IOC_IAX_CTIME |
-		      SCOUTFS_IOC_IAX_CRTIME | SCOUTFS_IOC_IAX_SIZE;
+	iax->x_mask = SCOUTFS_IOC_IAX_CTIME | SCOUTFS_IOC_IAX_CRTIME |
+		      SCOUTFS_IOC_IAX_SIZE;
 	iax->data_version = sm.data_version;
 	iax->ctime_sec = sm.ctime_sec;
 	iax->ctime_nsec = sm.ctime_nsec;
@@ -685,6 +685,9 @@ static long scoutfs_ioc_setattr_more(struct file *file, unsigned long arg)
 
 	if (sm.flags & SCOUTFS_IOC_SETATTR_MORE_OFFLINE)
 		iax->x_flags |= SCOUTFS_IOC_IAX_F_SIZE_OFFLINE;
+
+	if (sm.data_version != 0)
+		iax->x_mask |= SCOUTFS_IOC_IAX_DATA_VERSION;
 
 	ret = mnt_want_write_file(file);
 	if (ret < 0)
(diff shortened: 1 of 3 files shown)


versity-github:
Build 211 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/211/ git: aac84ebbf57e38f2940b375704810d399a0963c9

versity-github:
Build 187 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/187/ git: aac84ebbf57e38f2940b375704810d399a0963c9

kmod/src/ioctl.c
bendischinger, line 1: The comment at line 644 is no longer accurate: "We unconditionally fill the iax attributes from the sm set and let set_attr_x check them."

I'm wondering if it would make more sense to allow setting data_version = 0 if size = 0 at scoutfs/kmod/src/attr_x.c:132 since this seems to be a valid condition.

aversecat: A data version of `0` for a file that is being staged seems weird. Once there's data in it, it can't be zero anymore, data_version will always be `> 0`, right?

bendischinger: These are files that are being restored from a dump.  In the case of restoring an empty file, data version is 0 and size is 0.

aversecat:
Yes, but, do you actually release content with zero byte lengths? There are
no extents in a zero-byte file.

bendischinger:
> Yes, but, do you actually release content with zero byte lengths? There are no extents in a zero-byte file.

We may still make the call, not sure, would expect it to be a no op.

Any rate, I am fine with leaving the change as per Zach's comment in slack.
> So, really, just stop setattr_mroe from asking attr_x_set to change data_version if the setattr_more payload is 0.

bendischinger (approved):

aversecat:
retest

versity-github:
Build 229 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/229/ git: aac84ebbf57e38f2940b375704810d399a0963c9

versity-github:
Build 205 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/205/ git: aac84ebbf57e38f2940b375704810d399a0963c9

aversecat:
CI failed on `[  750.874382] scoutfs f.dc3622.r.2ff47d error: error -2 committing log merge: getting merge status item
`

zabbo (changes requested):
The commit title is correct, but the commit message and test language is misleading.  This has nothing to do with i_size.  It is only about the setattr_more wrapper setting the data_version flag while data_version is 0.  Which the code change correctly does.  Fix the commit summary and test to describe what's going on and we're off to the races.

aversecat:
I went all out on the wording to try and avoid the all the confusion. Hopefully this is much clearer this way now...

versity-github:
Build 234 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/234/ git: b45fbe0bbba45691258fe276c406569f9f8c1b8c

versity-github:
Build 210 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/210/ git: b45fbe0bbba45691258fe276c406569f9f8c1b8c

bendischinger (approved):

PR #184, opened 2024-09-12
Parallel_restore format v2 support

h-shaffer:
Add support for new format v2 structures. In this branch only scoutfs filesystem format v2 will be supported

### DIFF
--- a/tests/golden/parallel_restore
+++ b/tests/golden/parallel_restore
@@ -16,8 +16,10 @@ total_data_blocks 15728640
  ext:     logical_offset:        physical_offset: length:   expected: flags:
    0:        0..       0:          0..         0:      1:             last,unknown_loc,eof
     Type  Size     Total   Used      Free  Use%  
-MetaData  64KB    163840  34721    129119    21  
+MetaData  64KB    163840  34722    129118    21
     Data   4KB  15728640     64  15728576     0  
+== print test quota rule
+  7 13,L,- 15,L,- 17,L,- I 33 -
 == unmount small meta fs
 == just under ENOSPC
     Type  Size     Total    Used      Free  Use%
--- a/tests/golden/parallel_restore
+++ b/tests/golden/parallel_restore
@@ -25,4 +25,8 @@ MetaData  64KB    163840  163840         0   100
     Data   4KB  15728640      64  15728576     0
 == just over ENOSPC
 == ENOSPC
+== attempt to restore data device
+== parallel_restore data device
+== attempt format_v1 restore
+== parallel_restore format_v1 fs
 == cleanup
--- a/tests/golden/parallel_restore
+++ b/tests/golden/parallel_restore
@@ -29,4 +29,7 @@ MetaData  64KB    163840  163840         0   100
 == parallel_restore data device
 == attempt format_v1 restore
 == parallel_restore format_v1 fs
+== test if previously mounted
+== mount before restore
+== parallel_restore previously mounted device
 == cleanup
--- a/tests/src/parallel_restore.c
+++ b/tests/src/parallel_restore.c
@@ -601,7 +601,7 @@ static int do_restore(struct opts *opts)
 	ret = scoutfs_parallel_restore_create_writer(&wri);
 	error_exit(ret, "create writer %d", ret);
 
-	ret = scoutfs_parallel_restore_import_super(wri, super);
+	ret = scoutfs_parallel_restore_import_super(wri, super, dev_fd);
 	error_exit(ret, "import super %d", ret);
 
 	slices = calloc(1 + opts->nr_writers, sizeof(struct scoutfs_parallel_restore_slice));
--- a/tests/src/parallel_restore.c
+++ b/tests/src/parallel_restore.c
@@ -254,6 +254,33 @@ generate_entry(struct opts *opts, char *prefix, u64 nr, u64 dir_ino, u64 pos, u6
 	return entry;
 }
 
+/*
+ * since the _parallel_restore_quota_rule mimics the squota_rule found in the
+ * kernel we can also mimic its rule_to_irule function
+ */
+
+#define TEST_RULE_STR "7 13,L,- 15,L,- 17,L,- I 33 -"
+
+static struct scoutfs_parallel_restore_quota_rule *
+generate_quota(struct opts *opts)
+{
+	struct scoutfs_parallel_restore_quota_rule *prule;
+	int err;
+
+	prule = calloc(1, sizeof(struct scoutfs_parallel_restore_quota_rule));
+	error_exit(!prule, "Quota rule alloc failed");
+
+	err = sscanf(TEST_RULE_STR, " %hhu %llu,%c,%c %llu,%c,%c %llu,%c,%c %c %llu %c",
+		     &prule->prio,
+			 &prule->names[0].val, &prule->names[0].source, &prule->names[0].flags,
+		     &prule->names[1].val, &prule->names[1].source, &prule->names[1].flags,
+			 &prule->names[2].val, &prule->names[2].source, &prule->names[2].flags,
+			 &prule->op, &prule->limit, &prule->rule_flags);
+	error_exit(err != 13, "invalid quota rule, missing fields. nr fields: %d rule str: %s\n", err, TEST_RULE_STR);
+
+	return prule;
+}
+
 static u64 random64(void)
 {
 	return ((u64)lrand48() << 32) | lrand48();
@@ -564,6 +591,7 @@ static int do_restore(struct opts *opts)
 {
 	struct scoutfs_parallel_restore_writer *wri = NULL;
 	struct scoutfs_parallel_restore_slice *slices = NULL;
+	struct scoutfs_parallel_restore_quota_rule *rule = NULL;
 	struct scoutfs_super_block *super = NULL;
 	struct write_result res;
 	struct writer_args *args;
@@ -603,6 +631,11 @@ static int do_restore(struct opts *opts)
 
 	ret = scoutfs_parallel_restore_import_super(wri, super, dev_fd);
 	error_exit(ret, "import super %d", ret);
+
+	rule = generate_quota(opts);
+	ret = scoutfs_parallel_restore_add_quota_rule(wri, rule);
+	free(rule);
+	error_exit(ret, "add quotas %d", ret);
 
 	slices = calloc(1 + opts->nr_writers, sizeof(struct scoutfs_parallel_restore_slice));
 	error_exit(!slices, "alloc slices");
--- a/tests/tests/parallel_restore.sh
+++ b/tests/tests/parallel_restore.sh
@@ -10,7 +10,7 @@ mkdir -p "$SCR"
 # XXX this is all pretty manual, would be nice to have helpers
 echo "== make small meta fs"
 # meta device just big enough for reserves and the metadata we'll fill
-scoutfs mkfs -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
+scoutfs mkfs -V 2 -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
 	t_fail "mkfs failed"
 mount -t scoutfs -o metadev_path=$T_EX_META_DEV,quorum_slot_nr=0 \
 	"$T_EX_DATA_DEV" "$SCR"
@@ -60,6 +60,20 @@ echo "== ENOSPC"
 scoutfs mkfs -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
 	t_fail "mkfs failed"
 parallel_restore -m "$T_EX_META_DEV" -d 600:1000 -f 600:1000 -n 4000000 | grep died 2>&1 && t_fail "parallel_restore"
+
+echo "== attempt to restore data device"
+scoutfs mkfs -V 2 -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
+	t_fail "mkfs failed"
+
+echo "== parallel_restore data device"
+parallel_restore -m "$T_EX_DATA_DEV" | grep died 2>&1 && t_fail "parallel_restore"
+
+echo "== attempt format_v1 restore"
+scoutfs mkfs -V 1 -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
+	t_fail "mkfs failed"
+
+echo "== parallel_restore format_v1 fs"
+parallel_restore -m "$T_EX_META_DEV" | grep died 2>&1 && t_fail "parallel_restore"
 
 echo "== cleanup"
 rmdir "$SCR"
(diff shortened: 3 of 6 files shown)


versity-github:

aversecat (changes requested):
You wrote a few commit messages that are all in the title (first line) and thus overflowed into the body of the commit message.

You should always have a one-line (sub-72 character) title. Then an empty line, and then the elaborate commit message with s-o-b.

tests/tests/parallel_restore.sh
```diff
@@ -12,9 +12,6 @@ echo "== make small meta fs"
 # meta device just big enough for reserves and the metadata we'll fill
 scoutfs mkfs -V 2 -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
 	t_fail "mkfs failed"
-mount -t scoutfs -o metadev_path=$T_EX_META_DEV,quorum_slot_nr=0 \
-	"$T_EX_DATA_DEV" "$SCR"
-umount "$SCR"
```
aversecat, line 17: I have a separate commit to fix this bug - we should split it out because we can squash it away before it gets merged into `main` eventually.

utils/src/parallel_restore.c
```diff
@@ -1832,13 +1832,53 @@ spr_err_t scoutfs_parallel_restore_write_buf(struct scoutfs_parallel_restore_wri
...
+{
+	struct scoutfs_quorum_block *blk = NULL;
+	struct scoutfs_quorum_block_event *ev;
+	u64 blkno;
+	int i, j;
+	spr_err_t err;
+
+	for (i = 0; i <  SCOUTFS_QUORUM_MAX_SLOTS; i++) {
+		blkno = SCOUTFS_QUORUM_BLKNO + i;
+		free(blk);
```
aversecat, line 1911: Generally you shouldn't do this and the NULL assignment right before the allocation. Since read_block allocates, you want to start with blk = NULL (and you properly initialize it above), but because you free(blk) before read_block() is called, blk will only be freed if there are more iterations in the loop. On the last iteration, blk is never freed and you leak the memory.

In other words, just do read_block() ... free() every loop in that order, and make sure to account for error cases.

aversecat (changes requested):
You have a title here and the body is indented - probably because there is an empty line missing between them.

aversecat (changes requested):
Missing s-o-b here too...

utils/src/parallel_restore.h
```diff
@@ -46,6 +46,10 @@ struct scoutfs_parallel_restore_inode {
 	u32 gid;
 	u32 mode;
 	u32 rdev;
+#ifdef SCOUTFS_FORMAT_VERSION_FEAT_RETENTION
```
aversecat, line 49: I wouldn't do these #ifdefs here. These flags are universal for format V2 and they are going to be needed for all sorts of other flags in the future - they aren't tied to retention alone. Just like with the project ID we should always generate v2 content on disk when restoring.

So please just remove the ifdefs from this and the project ID patch, and if someone makes a wild version where it doesn't compile, they're on their own.

utils/src/parallel_restore.c
```diff
@@ -405,7 +405,10 @@ static spr_err_t map_start_key(struct scoutfs_key *start, struct scoutfs_key *ke
 		init_key(start, SCOUTFS_INODE_INDEX_ZONE, 0, 0,
 			 le64_to_cpu(key->_sk_second) & ~(u64)SCOUTFS_LOCK_SEQ_GROUP_MASK,
 			 0, 0);
-
+#ifdef SCOUTFS_FORMAT_VERSION_FEAT_QUOTA
```
aversecat, line 408: drop the ifdefs here, too

utils/src/parallel_restore.c
```diff
@@ -924,6 +927,48 @@ static spr_err_t insert_srch_item(struct scoutfs_parallel_restore_writer *wri,
 	return err;
 }
 
+#ifdef SCOUTFS_FORMAT_VERSION_FEAT_QUOTA
```
aversecat, line 930: drop this one too.

tests/src/parallel_restore.c
```diff
@@ -254,6 +256,31 @@ generate_entry(struct opts *opts, char *prefix, u64 nr, u64 dir_ino, u64 pos, u6
 	return entry;
 }
 
+/*
+ * since the _parallel_restore_quota_rule mimics the squota_rule found in the
+ * kernel we can also mimic its rule_to_irule function
+ */
+
+#define DEFAULT_RULE_STR "7 13,L,- 15,L,- 17,L,- I 33 -"
```
aversecat, line 264: Maybe call it TEST_RULE_STR instead?

tests/src/parallel_restore.c
```diff
@@ -254,6 +256,31 @@ generate_entry(struct opts *opts, char *prefix, u64 nr, u64 dir_ino, u64 pos, u6
...
+
+static struct scoutfs_parallel_restore_quota_rule *
+generate_quotas(struct opts *opts)
+{
+	struct scoutfs_parallel_restore_quota_rule *prule;
+	int err;
+
+	prule = calloc(1,sizeof(struct scoutfs_parallel_restore_quota_rule));
```
aversecat, line 272: space after comma

tests/src/parallel_restore.c
```diff
@@ -596,14 +623,17 @@ static int do_restore(struct opts *opts)
 
 	ret = pread(dev_fd, super, SCOUTFS_BLOCK_SM_SIZE,
 		    SCOUTFS_SUPER_BLKNO << SCOUTFS_BLOCK_SM_SHIFT);
-	error_exit(ret != SCOUTFS_BLOCK_SM_SIZE, "error reading super, ret %d", ret);
+	error_exit(ret != SCOUTFS_BLOCK_SM_SIZE, "errorreading super, ret %d", ret);
```
aversecat, line 626: probably an editor accident

tests/src/parallel_restore.c
```diff
@@ -254,6 +256,31 @@ generate_entry(struct opts *opts, char *prefix, u64 nr, u64 dir_ino, u64 pos, u6
...
 
+/*
+ * since the _parallel_restore_quota_rule mimics the squota_rule found in the
+ * kernel we can also mimic its rule_to_irule function
+ */
+
+#define DEFAULT_RULE_STR "7 13,L,- 15,L,- 17,L,- I 33 -"
+
+static struct scoutfs_parallel_restore_quota_rule *
+generate_quotas(struct opts *opts)
```
aversecat, line 267: the function here generates a single quota rule. The name should reflect this.

tests/golden/parallel_restore
```diff
@@ -16,7 +16,7 @@ total_data_blocks 15728640
  ext:     logical_offset:        physical_offset: length:   expected: flags:
    0:        0..       0:          0..         0:      1:             last,unknown_loc,eof
     Type  Size     Total   Used      Free  Use%  
-MetaData  64KB    163840  34721    129119    21  
+MetaData  64KB    163840  34722    129118    21
```
aversecat, line 19: You should verify that we can actually see that the quota rule was properly added by listing the quota rules for the filesystem here.

tests/src/parallel_restore.c
```diff
@@ -596,14 +623,17 @@ static int do_restore(struct opts *opts)
...
 
 	ret = scoutfs_parallel_restore_create_writer(&wri);
 	error_exit(ret, "create writer %d", ret);
 
-	ret = scoutfs_parallel_restore_import_super(wri, super);
+	ret = scoutfs_parallel_restore_import_super(wri, super, dev_fd);
 	error_exit(ret, "import super %d", ret);
 
+	ret = scoutfs_parallel_restore_add_quota_rule(wri, generate_quotas(opts));
+	error_exit(ret, "add quotas %d", ret);
```
aversecat, line 635: Another thing to add is to implement quota restore in `restore_copy.c`.

aversecat (changes requested):

tests/tests/parallel_restore.sh
h-shaffer: Hmm, github isn't making it too clear what you're referencing here, could you elaborate a bit?

utils/src/parallel_restore.h
h-shaffer: The original thought for this was that these constants were guaranteed to exist on a format v2 fs where they wouldn't on fmt < v2. However, I agree they should be removed now that we only allow version v2.

aversecat:
You are deleting these lines because a different test shows that they were
added by mistake. The problem is that the removal should be a separate
commit explaining why they need to be removed.

I have a patch for this already. You can redo yours and make it into a
separate commit, or just drop it.

tests/src/parallel_restore.c
h-shaffer: Cleaned up everything else, I'm going to push those changes up while I work on this

utils/src/parallel_restore.h
```diff
@@ -68,6 +68,21 @@ struct scoutfs_parallel_restore_inode {
...
+struct scoutfs_parallel_restore_quota_rule {
+	u64 limit;
+	u8  prio;
+	u8  op;
+	u8  rule_flags;
+	struct quota_rule_name {
+		u64 val;
+		u8  source;
+		u8  flags;
+		u8  pad[7];
```
aversecat, line 80: This should be a different length padding, don't you agree? Moreover, `names` isn't aligned at all right now, so we'd want another pad in front to make that align. But I really wonder if you just can't use `struct scoutfs_ioctl_quota_rule` instead, which has all the correct padding.

utils/src/parallel_restore.c
```diff
@@ -1849,7 +1899,7 @@ static spr_err_t scoutfs_check_if_previous_mount(int fd)
 	for (i = 0; i <  SCOUTFS_QUORUM_MAX_SLOTS; i++) {
 		blkno = SCOUTFS_QUORUM_BLKNO + i;
 		err = read_block(fd, blkno, SCOUTFS_BLOCK_SM_SHIFT, (void **)&blk);
-		if (blk == NULL;)
+		if (blk == NULL)
```
aversecat, line 1902: Don't fix up unrelated bugs without any comments like this. Make them a separate commit.

tests/tests/parallel_restore.sh
```diff
@@ -61,6 +61,45 @@ scoutfs mkfs -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DA
 	t_fail "mkfs failed"
 parallel_restore -m "$T_EX_META_DEV" -d 600:1000 -f 600:1000 -n 4000000 | grep died 2>&1 && t_fail "parallel_restore"
 
+echo "== just under ENOSPC"
```
aversecat, line 67: This is duplicating a bunch of existing tests. merge issue?

utils/src/parallel_restore.c
```diff
@@ -1832,13 +1832,55 @@ spr_err_t scoutfs_parallel_restore_write_buf(struct scoutfs_parallel_restore_wri
...
+	struct scoutfs_quorum_block *blk = NULL;
+	struct scoutfs_quorum_block_event *ev;
+	u64 blkno;
+	int i, j;
+	spr_err_t err;
+
+	for (i = 0; i <  SCOUTFS_QUORUM_MAX_SLOTS; i++) {
+		blkno = SCOUTFS_QUORUM_BLKNO + i;
+		err = read_block(fd, blkno, SCOUTFS_BLOCK_SM_SHIFT, (void **)&blk);
+		if (blk == NULL;)
```
aversecat: still has the semicolon here... please remove it.

or just write `if (!blk)`

utils/src/parallel_restore.c
```diff
@@ -1832,13 +1832,55 @@ spr_err_t scoutfs_parallel_restore_write_buf(struct scoutfs_parallel_restore_wri
...
+			return EINVAL;
+
+		free(blk);
```
aversecat, line 1905: You can't free it here yet. You're reading its contents 6 lines below.

utils/src/parallel_restore.c
```diff
@@ -1873,6 +1915,7 @@ spr_err_t scoutfs_parallel_restore_import_super(struct scoutfs_parallel_restore_
 spr_err_t scoutfs_parallel_restore_export_super(struct scoutfs_parallel_restore_writer *wri,
 						struct scoutfs_super_block *super)
 {
+
```
aversecat, line 1970: remove this.

tests/src/parallel_restore.c
```diff
@@ -604,6 +631,9 @@ static int do_restore(struct opts *opts)
 	ret = scoutfs_parallel_restore_import_super(wri, super, dev_fd);
 	error_exit(ret, "import super %d", ret);
 
+	ret = scoutfs_parallel_restore_add_quota_rule(wri, generate_quota(opts));
```
aversecat, line 634: `generate_quota` allocates memory, and returns a pointer to it. The caller should then free this memory after it's been used.

aversecat (changes requested):

utils/src/parallel_restore.c
h-shaffer: Hmm I was pretty certain that this got fixed in my changes but it doesn't seem so, I'll make sure its fixed in the next push

utils/src/parallel_restore.c
```diff
@@ -1781,6 +1822,22 @@ spr_err_t scoutfs_parallel_restore_add_progress(struct scoutfs_parallel_restore_
...
+	spr_err_t err;
+	if (!wri_has_super(wri)) {
+		err = EINVAL;
+		goto out;
+	}
+
+	err = insert_quota_item(wri, rule);
+out:
+	if (rule)
+		free(rule);
```
aversecat, line 1837: No, I think this is bad API design. The library should stay consistent and not free pointers here. The data here could have been on the stack, or part of some other struct that the caller maintains, like an array.

The caller should free the `rule` after `scoutfs_parallel_restore_add_quota_rule()` returns. Ths same way that `scoutfs_parallel_restore_add_entry` and `scoutfs_parallel_restore_add_inode` also do not free the pointer.

In general, if the library allocates the memory, it should by default also free() it when it is appropriate. For stuff that's just passed in by pointer, it should by default not free the memory.

utils/src/parallel_restore.c
```diff
@@ -818,6 +818,7 @@ static spr_err_t insert_inode_items(struct scoutfs_parallel_restore_writer *wri,
 	si->mode = cpu_to_le32(inode->mode);
 	si->rdev = cpu_to_le32(inode->rdev);
 	si->flags = 0;
+	si->flags = cpu_to_le64(inode->flags);
```
aversecat, line 822: Can we have a quick test case to make sure this works?

utils/src/parallel_restore.c
```diff
@@ -826,6 +826,7 @@ static spr_err_t insert_inode_items(struct scoutfs_parallel_restore_writer *wri,
 	si->mtime.nsec = cpu_to_le32(inode->mtime.tv_nsec);
 	si->crtime.sec = cpu_to_le64(inode->crtime.tv_sec);
 	si->crtime.nsec = cpu_to_le32(inode->crtime.tv_nsec);
+	si->proj = cpu_to_le64(inode->proj);
```
aversecat, line 831: Same here, we should have a test that validates that this works.

aversecat (changes requested):
Just one change and some extra tests needed I think now.

tests/tests/parallel_restore.sh
```diff
@@ -61,6 +61,20 @@ scoutfs mkfs -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DA
 	t_fail "mkfs failed"
 parallel_restore -m "$T_EX_META_DEV" -d 600:1000 -f 600:1000 -n 4000000 | grep died 2>&1 && t_fail "parallel_restore"
 
+echo "== attempt to restore data device"
+scoutfs mkfs -V 2 -A -f -Q 0,127.0.0.1,53000 -m 10G -d 60G "$T_EX_META_DEV" "$T_EX_DATA_DEV" > $T_TMP.mkfs.out 2>&1 || \
+	t_fail "mkfs failed"
+
+echo "== parallel_restore data device"
```
aversecat, line 71: The test output of these things ended up in the next patch by accident.

aversecat:
Obsolete now, but we should make sure we revisit the outstanding items against the other branch.

PR #188, opened 2024-09-19
Allow skipping of tests.

aversecat:
With el9, we will be skipping the format-version-forward-back test on el9 platforms, because the test requires that we compile and run code that has no el9 support. For that test, we will rely on older OS versions to not fail, and it will not be skipped there.

So we have to modify our outcome checks otherwise all el9 builds will fail testing no matter what. This is the simplest form where tests can skip - only failing tests will return an error code.

### DIFF
--- a/tests/run-tests.sh
+++ b/tests/run-tests.sh
@@ -644,16 +644,16 @@ if [ -n "$T_TRACE_GLOB" -o -n "$T_TRACE_PRINTK" ]; then
 	fi
 fi
 
-if [ "$skipped" == 0 -a "$failed" == 0 ]; then
-	msg "all tests passed"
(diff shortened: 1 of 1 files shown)


versity-github:
Build 223 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr/223/ git: c148c3c56d03ffa20385a603c85a6fa5aaa5f5f9

versity-github:
Build 199 http://jenkins.vpn.versity.com:8080/job/scoutfs-run-pr-el8/199/ git: c148c3c56d03ffa20385a603c85a6fa5aaa5f5f9

aversecat:
Modified and moved into `auke/el9` as part of #151

PR #190, opened 2024-10-10
Don't fence clients with rid==0

aversecat:
Occasionally during the export-lookup-evict-race test we see the following failure dmesg output when a server fences a client that has no valid rid:

[  828.379546] sysfs: cannot create duplicate filename '/fs/scoutfs/f.b928e1.r.7b36c0/fence/0000000000000000' ...
[  828.379773] kobject_add_internal failed for 0000000000000000 with -EEXIST, don't try to register things with the same name in the same directory. [  828.385946] scoutfs f.b928e1.r.7b36c0 error: client fence returned err -17, shutting down server

This fails the test. Fencing these clients is unwanted, but we definitely don't want to create duplicate sysfs entries for it, either.

Don't fence clients like this, just return 0.

### DIFF
--- a/kmod/src/fence.c
+++ b/kmod/src/fence.c
@@ -238,6 +238,11 @@ int scoutfs_fence_start(struct super_block *sb, u64 rid, __be32 ipv4_addr, int r
 	struct pending_fence *fence;
 	int ret;
 
+	if (!rid) {
+		ret = 0;
+		goto out;
+	}
+
 	fence = kzalloc(sizeof(struct pending_fence), GFP_NOFS);
 	if (!fence) {
 		ret = -ENOMEM;


aversecat:
```
[  828.379546] sysfs: cannot create duplicate filename '/fs/scoutfs/f.b928e1.r.7b36c0/fence/0000000000000000'
[  828.379548] CPU: 1 PID: 14628 Comm: kworker/u4:4 Kdump: loaded Tainted: G           OE    --------- -  - 4.18.0-513.9.1.el8_9.x86_64 #1
[  828.379550] Hardware name: Red Hat KVM, BIOS 1.16.0-4.module+el8.9.0+1408+7b966129 04/01/2014
[  828.379552] Workqueue: scoutfs_net_server scoutfs_net_reconn_free_worker [scoutfs]
[  828.379582] Call Trace:
[  828.379607]  dump_stack+0x41/0x60
[  828.379612]  sysfs_warn_dup.cold.4+0x17/0x2a
[  828.379617]  sysfs_create_dir_ns+0xb3/0xe0
[  828.379620]  kobject_add_internal+0xc2/0x290
[  828.379622]  kobject_init_and_add+0x71/0xa0
[  828.379624]  scoutfs_sysfs_create_attrs_parent+0x16c/0x1a0 [scoutfs]
[  828.379658]  ? scoutfs_fence_start+0x43/0x190 [scoutfs]
[  828.379688]  ? kmem_cache_alloc_trace+0x142/0x280
[  828.379691]  scoutfs_fence_start+0xb4/0x190 [scoutfs]
[  828.379711]  scoutfs_net_reconn_free_worker+0x1bd/0x270 [scoutfs]
[  828.379732]  process_one_work+0x1d3/0x390
[  828.379736]  ? process_one_work+0x390/0x390
[  828.379738]  worker_thread+0x30/0x390
[  828.379740]  ? process_one_work+0x390/0x390
[  828.379742]  kthread+0x134/0x150
[  828.379745]  ? set_kthread_struct+0x50/0x50
[  828.379747]  ret_from_fork+0x35/0x40
[  828.379773] kobject_add_internal failed for 0000000000000000 with -EEXIST, don't try to register things with the same name in the same directory.
[  828.385946] scoutfs f.b928e1.r.7b36c0 error: client fence returned err -17, shutting down server
export-lookup-evict-race unexpected messages in dmesg
```

for reference - caller path is known.

kmod/src/fence.c
```diff
@@ -238,6 +238,11 @@ int scoutfs_fence_start(struct super_block *sb, u64 rid, __be32 ipv4_addr, int r
 	struct pending_fence *fence;
 	int ret;
 
+	if (!rid) {
```
zabbo, line 241: Sadly, I don't think this'll work.  Depending on where the fence came from we might not end up cleaning it up.  See the _fence_next calls in the server that lead to _recov_finished.  We'll need to more about the precise source of the fencing request to avoid it.

zabbo (changes requested):

kmod/src/fence.c
aversecat: So, we know it comes from `scoutfs_net_reconn_free_worker+0x1bd` and not the other 2 callers. I'll look into that.

Also, we can just stopgap and not create sysfs entries here, too, and just let all the fencing happen. The worst that could happen, I guess, is that we end up fencing the wrong clients that also have rid=0, so maybe not such a big deal?

PR #191, opened 2024-10-16
Restore and Check.

aversecat:
This branch against `main` contains both the parallel restore library, tests, as well as the check code.

The check code has been reduced to our validation tool to identify inconsistencies in an umounted scoutfs meta device, only. All error injection and repair code has been omitted. The check code is used to validate the output of the parallel restore test cases.

- [x] - tests need to pass, but check fails. There either is a block reading problem in `scoutfs check` or there is an error inserting during `restore`. 
- [x] - quota print outputs incorrect quota rules during test

New issues from @chaowang-versity:

- [ ] - parallel_restore binary fails with more than 10 threads
- [x] - hardlinks are incorrect: After parallel_restore, hardlinked items have a hardlink count of `1` where that number should be larger than `1` for obvious reasons.
- [ ] - `util.h` should be exported, because `format.h` uses `DIV_ROUND_UP` which is defined in it.


aversecat:
pushed el7 fixes into this.

zabbo:
Let's pull out the supporting commits that are fine on their own and don't really have much to do with parallel restore.  We can land them in their own PR and then focus on fixing up the parallel restore commits.  I think the following should do it: ```bb2003cf Fix printing alloc list block extents
0d156e02 Import a few more functions to our list.h
3f3c2d86 Add userspace NSEC_PER_SEC
9cbe0411 Add bloom filter index calc for userspace utils
fcdefb72 Add srch_encode_entry() for userspace utils
b3a83802 Add put_unaligned_leXX() for userspace
ad8a0f7e Add fls64() alias for userspace flsll()
9524397d Promote userspace btree block initialization
cd43c391 Add userspace version of our mode to type
29ca2ad2 Add userspace version of our dirent name hash
0020eac0 Add lk rbtree wrapper
1f7f40c9 Add test_bit to utils bitmap```

aversecat:
See: #201

aversecat:
rebased onto `main`

aversecat:
This is failing tests on `scoutfs df` output. We'll need to scrub it or always force meta/data sizing, otherwise it'll always diff.

zabbo:
> This is failing tests on `scoutfs df` output. We'll need to scrub it or always force meta/data sizing, otherwise it'll always diff.

Yeah, the test shouldn't be letting the raw df output through to be compared.

aversecat:
rebased onto main. added nlink fix. Still needs `df` test output scrub fix.

PR #192, opened 2024-10-24
Generate debug packages

gcymbalski:
We had previously explicitly disabled this; let's start generating them.


gcymbalski:
For reference, the raw build jobs did in fact spit out a debuginfo package: http://jenkins.vpn.versity.com:8080/job/ng-scoutfs-build-el7/260/artifact/kmod/scoutfs_7_9/

aversecat (approved):

PR #193, opened 2024-10-31
Zab/hung lock fixes

zabbo:


zabbo:
The el7 failure is an unrelated instability in one test, everything else ran fine.  [ 1572.077909] scoutfs f.7a8659.r.b1166d error: error -2 committing log merge: getting merge status item
mount-unmount-race unexpected messages in dmesg

tests/tests/lock-shrink-read-race.sh
```diff
@@ -0,0 +1,38 @@
+#
+# We had a lock server refcounting bug that could let one thread get a
+# reference on a lock struct that was being freed by another thread.  We
+# were able to reproduce this by having all clients try and produce a
+# lot of read and null requests.
+#
+# This will manfiest as a hung lock and timed out test runs, probably
+# with hung task messages on the console.
```
aversecat, line 8: I would add the comment that it will trigger KASAN when it's enabled. Perhaps with a fingerprint of the KASAN output.

kmod/src/net.c
```diff
@@ -519,20 +519,20 @@ static int move_acked_responses(struct scoutfs_net_connection *conn,
...
 /* acks are processed inline in the recv worker */
 static void free_acked_responses(struct scoutfs_net_connection *conn, u64 seq)
 {
-	int moved;
+	bool moved;
 
 	spin_lock(&conn->lock);
 
-	moved = move_acked_responses(conn, &conn->send_queue, seq) +
+	moved = move_acked_responses(conn, &conn->send_queue, seq) |
```
bendischinger, line 535: Just wondering if this is just to clean up some code, looks like before and after are logically the same.  I suppose if the ints could have been large they could've happened to overflow back to zero, but not here since the only possible values were 0 and 1.

zabbo: Yup, just code cleanup only.  It looked weird enough, but the change is safe enough, that I'm not worried about including it :).

bendischinger (approved):
Fix makes sense to me and good explanations for all of the commit descriptions.

tests/tests/lock-shrink-read-race.sh
zabbo: Fair enough.  I added a note, though it's not guaranteed, and mentioned the function it'll be found in.  (also fixed some silly trailing whitespace in the golden output).

aversecat (commented):
sp: "betwen" in the commit message.

aversecat (approved):

PR #195, opened 2024-11-09
RHEL9.5 kernel support

aversecat:
3 patches for rhel9.5. Kicking off for CI builds. Still want to make the one patch a little less ugly.

### DIFF
--- a/kmod/src/Makefile.kernelcompat
+++ b/kmod/src/Makefile.kernelcompat
@@ -422,3 +422,12 @@ endif
 ifneq (,$(shell grep 'struct backing_dev_info.*backing_dev_info;' include/linux/sched.h))
 ccflags-y += -DKC_CURRENT_BACKING_DEV_INFO
 endif
+
+#
+# v6.8-rc1-4-gf3a608827d1f
+#
+# adds bdev_file_open_by_path() and later in v6.8-rc1-30-ge97d06a46526 removes bdev_open_by_path()
+# which requires us to use the file method from now on.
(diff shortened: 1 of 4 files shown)


aversecat:
retest

zabbo:
The patches look fine, but have we looked at the generic/604 xfstests failure in 9.5?  It looks like it's related to moun/unmount so the file bdev change might be related.

aversecat:
Oh, I hadn't actually seen that failure in CI until just now. I will definitely look at it - unsure why I'm not seeing this in my own testing!

aversecat:
retest

aversecat:
`vanilla-el94` failed with:

```
[ 2280.583161] scoutfs f.b92def.r.d4b0a7 error: error -5 getting merge req rid 32afdaeb82bbb15c: finding next logged item
```

This has been creeping in a little here and there. I don't think this is related to any of the patches in this series, but we should def. look into it.

aversecat:
Still more intermittent failures but nothing new (`large-fragmented-free` and `orphan-inodes`).

zabbo:
> `vanilla-el94` failed `quorum-heartbeat-timeout` with dmesg output:
> 
> ```
> [ 2280.583161] scoutfs f.b92def.r.d4b0a7 error: error -5 getting merge req rid 32afdaeb82bbb15c: finding next logged item
> ```
> 
> This has been creeping in a little here and there. I don't think this is related to any of the patches in this series, but we should def. look into it.

Thanks for taking a look.  I agree that it's unrelated to the patches.  I think this is EIO escaping from forced unmount.  It would be nice to find a clean way to avoid emitting errors when we know we're tearing the world down destructively, but I also don't want to get too lost in the problem.

PR #196, opened 2024-11-14
Make `df` print out blocks free based on data, only

aversecat:
The current `df` output displays an estimate of free disk space available based on the usage of both meta and data device.

Change it to purely report back on data device usage only.

There are arguments both for and against this. First, it is an obvious change in current behavior, that may break existing scripts and usage reporting methods.

Arguments for this are that this is easier to understand for users - the number will now reflect whether more data can be stored, or whether data needs to be released or not to meet availability requirements.

### DIFF
--- a/kmod/src/super.c
+++ b/kmod/src/super.c
@@ -102,12 +102,10 @@ static int scoutfs_statfs(struct dentry *dentry, struct kstatfs *kst)
 	if (ret)
 		goto out;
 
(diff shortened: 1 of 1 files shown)


zabbo:
I don't think we should make this change.  There is the con that we'd be changing it's behaviour, and that might be enough, but also consider that it'd then work differently than everything else.  df simply isn't universally predictive of allocation success and it hasn't been for a long time.  (btrfs, thin provisioning, superuser reservations -- lots of structures operate counter to what df might indicate).  There's also the observation that if we stop including metadata in df then metadata allocations could give enospc while df looked fine again.  It's just the wrong tool.

PR #197, opened 2024-12-02
Block EL9 minor version upgrades

gcymbalski:
Since kABI migrations across minor versions is a thing of the past going
forward, we now:
- Detect if we're on EL9
- If so, add a requirement on the various flavors of release package to
  that specific major.minor version

This appropriately does not allow upgrades across minor versions.


aversecat (approved):
OK, yeah, I can see this work. I assume there isn't an easier way to do this without falling back on lua, and that this works across all versions of RHEL?!

aversecat:
Needs retesting though, -vanilla-el9* should pass.

PR #200, opened 2025-01-10
Draft for offline restore

chaowang-versity:
This is a preview of the offline restore utility from Golang side.  There are still things needed to decide like which repo this code should go, how to proper test the code, how to handle / log errors.


tests/src/restore_copy.c
```diff
@@ -27,6 +27,9 @@
 #include "../../kmod/src/ioctl.h"
 #include "../../utils/src/parallel_restore.h"
 
+#include <zlib.h>
```
aversecat, line 30: You can include "crc.h" instead here.

tests/src/restore_copy.c
```diff
@@ -897,6 +911,9 @@ static int do_restore(struct opts *opts)
 	ret = scoutfs_parallel_restore_export_super(wri, super);
 	error_exit(ret, "update super %d", ret);
 
+
+	fprintf(stdout, "writing Super block signature: %08x\n",
```
aversecat, line 915: I'm not opposed to making this a little more verbose, as it's a test binary anyway, but it's a little inconsistent with mixing printf/fprintf and indentation.

I think we can just leave this out once done debugging the initial problems, though.

tests/Makefile
```diff
@@ -19,7 +19,7 @@ BIN := src/createmany			\
 
 DEPS := $(wildcard src/*.d)
 
-all: $(BIN)
+ll: $(BIN)
```
aversecat, line 22: typo I think

tests/Makefile
```diff
@@ -1,4 +1,4 @@
-CFLAGS := -Wall -O2 -Werror -D_FILE_OFFSET_BITS=64 -fno-strict-aliasing -I ../kmod/src
+CFLAGS := -Wall -O2 -D_FILE_OFFSET_BITS=64 -fno-strict-aliasing -I ../kmod/src
```
aversecat, line 1: only for testing PRs, this can't stay in here.

restore/pkg/restore/restore.go
```diff
@@ -0,0 +1,878 @@
+package restore
+
+/*
+#cgo CFLAGS: -I${SRCDIR}/../../../utils/src -I${SRCDIR}/../../../kmod/src
+#cgo LDFLAGS: -L${SRCDIR}/../../../utils/src -l:scoutfs_parallel_restore.a -lm
```
aversecat, line 5: I think this is going to be the crux when we decide whether to make this its own git tree, or keep it in here:

- You're coding these cgo flags to include stuff directly from the source tree, and not from the installed headers
- You're pulling in internal header files that aren't currently "exported" (util.h, format.h, sparse.h).

There's good reasons why you're doing this - I'm not saying it is wrong.

I tend to feel like I want to keep all the `go` code out of `scoutfs` as is. I certainly don't want to have normal scoutfs builds and test runs have to install the go toolchain. It would likely cost significant extra time to our test runs.

The other side is that we can just have this in the scoutfs tree and as long as nothing hooks into `make` it will just be excluded/ignored.

Additional thoughts:
- having some actual test/validation code
- how about a "test" database that we can use to validate this?
- some quick build instructions
- maybe push this into a `go/restore` subfolder instead. `restore` as toplevel seems weird. Even `utils/go/resore` seems better.

tests/src/restore_copy.c
chaowang-versity: Thos are just for debugging purposes, I did not mean to change this file and Makefile. I will revert  them

tests/Makefile
chaowang-versity: I will revert the restore_copy.c and Makefile

tests/Makefile
chaowang-versity: I will revert the restore_copy.c and Makefile

restore/pkg/restore/restore.go
chaowang-versity: Thanks for the feedback. One of the reasons I posted this PR as a preview is to facilitate the discussion about where this code should go.  I also would like to learn how scoutfs team does testing, as it is a user space utility, it needs an environment that it can create and mount a scoutfs fs before doing any validation.

chaowang-versity:
moving to https://github.com/versity/fastrestore

PR #202, opened 2025-02-11
Swap meta allocators with much more reserve.

aversecat:
This is the core of the meta allocator swapping change: swap a little bit earlier, but always swap only when _freed is much larger than _avail. 

From my observations I can see that without this change, both allocators bleed slowly to empty without either of them regaining enough over time to continue work. In effect, we never recover from ENOSPC. Everything grinds to a halt, the allocators just swap endlessly and transactions take forever to complete.

With this change we avoid all that - we purposely let our _avail allocator go much lower than _freed before we swap. This assures that whenever we swap we let enough blocks recycle post transactions into _freed so that it gets us out of ENOSPC again, and into a healthy state again.

I think.

### DIFF
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -668,14 +668,16 @@ static void scoutfs_server_commit_func(struct work_struct *work)
 	 * the reserved blocks after having filled the log trees's avail
 	 * allocator during its transaction.  To avoid prematurely
 	 * setting the low flag and causing enospc we make sure that the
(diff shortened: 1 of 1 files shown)


aversecat:
retest

aversecat:
retest

aversecat:
retest

aversecat:
retest

aversecat:
retest

aversecat:
Ironically, `enospc` failed on el7 vanilla, in the classic failure pattern.

PR #203, opened 2025-02-18
Fix new_inode ctime assignment.

aversecat:
Very old copy/paste bug here, we want to update new_inode's ctime instead. old_inode already is updated.


zabbo (approved):
Heh, that is indeed ancient :).  Good catch.

PR #205, opened 2025-03-05
Random small fixes.

aversecat:


aversecat:
retest

aversecat:
I've taken out the nonsense commits. What remains is:
- two ioctl data leaks, plugged by `memset`
- a memory leak in mounting (should free super if errors occur)
- two error paths that will OOPS on unlikely failure

seems reasonable enough despite the low chance that these cause any actual issues.

kmod/src/ioctl.c
```diff
@@ -1373,8 +1376,7 @@ static long scoutfs_ioc_get_referring_entries(struct file *file, unsigned long a
 			ent.name_len = name_len;
 
 			if (copy_to_user(uent, &ent, sizeof(struct scoutfs_ioctl_dirent)) ||
-			    copy_to_user(&uent->name[0], bref->dent.name, name_len) ||
-			    put_user('\0', &uent->name[name_len])) {
```
zabbo, line 1374: This code dates back from the early times when the _dirent struct was __packed.  Now that it's padded, this all doesn't make much sense.  We still need the put_user for the null term because name_len doesn't include the null.  But it's weird that we copy the whole struct and then overwrite the name.  The first copy should stop at the offset of the name, not copy the whole size.

kmod/src/super.c
```diff
@@ -511,10 +511,10 @@ static int scoutfs_fill_super(struct super_block *sb, void *data, int silent)
 	mapping_set_gfp_mask(sb->s_bdev->bd_inode->i_mapping, GFP_NOFS);
 
 	sbi = kzalloc(sizeof(struct scoutfs_sb_info), GFP_KERNEL);
-	sb->s_fs_info = sbi;
```
zabbo, line 514: Careful.  that's a load bearing sb->s_fs_info = NULL.  It affects cleanup paths that look to see if s_fs_info was built up.  Leave it, and move the sb->s_fs_info below.

kmod/src/srch.c
```diff
@@ -2254,10 +2254,9 @@ static void scoutfs_srch_compact_worker(struct work_struct *work)
 	int err;
 
 	sc = kmalloc(sizeof(struct scoutfs_srch_compact), GFP_NOFS);
-	if (sc == NULL) {
-		ret = -ENOMEM;
-		goto out;
-	}
+	if (sc == NULL)
+		/* don't goto out, we'd deref sc->nr! */
+		return;
```
zabbo, line 2259: This just shut down compaction forever on the mount.  The boolean being passed into the queue impacts how long again before it'll retry.  If there's an error, including this, it should try again later.  Just add sc != NULL && to the condition.

zabbo (changes requested):

kmod/src/srch.c
aversecat: yeahhh, makes total sense. ack.

kmod/src/ioctl.c
aversecat: Call me silly, but, isn't this a much less complicated solution here:
```
diff --git a/kmod/src/ioctl.c b/kmod/src/ioctl.c
index 432e6d37..f05dad0d 100644
--- a/kmod/src/ioctl.c
+++ b/kmod/src/ioctl.c
@@ -1371,10 +1371,10 @@ static long scoutfs_ioc_get_referring_entries(struct file *file, unsigned long a
                        ent.flags = bref->last ? SCOUTFS_IOCTL_DIRENT_FLAG_LAST : 0;
                        ent.d_type = bref->d_type;
                        ent.name_len = name_len;
+                       ent.name[0] = ent.name[1] = ent.name[2] = 0;
+                       memcpy(&ent.name[0], bref->dent.name, name_len);
 
-                       if (copy_to_user(uent, &ent, sizeof(struct scoutfs_ioctl_dirent)) ||
-                           copy_to_user(&uent->name[0], bref->dent.name, name_len) ||
-                           put_user('\0', &uent->name[name_len])) {
+                       if (copy_to_user(uent, &ent, sizeof(struct scoutfs_ioctl_dirent))) {
                                ret = -EFAULT;
                                goto out;
                        }
```

We can avoid the memset(). We're doing a full copy_to_user() of 2 compatible structs, so no weird split copy. We're zeroing 3 u8's prior to assure we're not leaking and null-terminating the string. And ent.name is properly filled.

aversecat: nvm, forget this. I see now.

aversecat: This one is sorted out as well.

zabbo:
retest

PR #207, opened 2025-03-12
Avoid null deref while walking avl tree

aversecat:
Avoid a possible null dereference here by checking node is always valid before walking ->left or ->right.


zabbo:
We don't think this is valid.  The caller tests that ->left/right are non-null before calling.  This was done in response to a gcc -fanalyzer run, and it probably exceeded its call chain context and didn't see the callers constraints.

PR #216, opened 2025-04-29
Zab/stop ending dirty data freed

zabbo:


zabbo:
The el8 failure was a spurious back-compat git repo failure.. the rest of the tests were fine.  I'm landing this so we can get some time with it in the periodic runners on main.

PR #217, opened 2025-05-05
Do not fence connections without valid greeting. (replaces #190)

aversecat:
There is no reason to fence any connection that hasn't sent a valid greeting, since they haven't progressed far enough for it to make sense.

This prevents the server from emitting:
```
    client 127.0.0.1:12345 reconnect timed out, fencing
```

~~But, it doesn't actually affect the behavior much. No fencing is taking place at this time. The server will still close all listening sockets and hand over to the next server, and all clients will disconnect and reconnect to the new leader.~~

### DIFF
--- a/kmod/src/net.c
+++ b/kmod/src/net.c
@@ -1452,15 +1452,23 @@ restart:
 			set_conn_fl(acc, reconn_freeing);
 			spin_unlock(&conn->lock);
 			if (!test_conn_fl(conn, shutting_down)) {
-				scoutfs_info(sb, "client "SIN_FMT" reconnect timed out, fencing",
-					     SIN_ARG(&acc->last_peername));
-				ret = scoutfs_fence_start(sb, acc->rid,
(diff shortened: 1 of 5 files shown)


aversecat:
Replaces #190.

The underlying problem is that there seems to be nothing in place to drop outdated connections from the server, aside disconnecting all clients, restarting it, and having everyone connect to the new server. If this would be added, then the server wouldn't have to do this interruptive process of restarting the listening socket.

So, this is just a start to address the issue.

aversecat:
retest

aversecat:
Rebased. The complex half of the patch completely dropped out because we implemented it in `v1.33` already. So now it's just the easy half.

aversecat:
retest

zabbo (changes requested):
for posterity: this short-circuited fencing pretty far down the line in the connection life time.  It still printed messages to the console, waited for a reconnection timeout, and most notably still would call the notify_down callback with a 0 rid which would shut down the server.

PR #218, opened 2025-05-05
Clk/cluster lock spinlock

chriskirby1:


kmod/src/lock.c
```diff
@@ -1454,44 +1645,53 @@ static unsigned long lock_scan_objects(struct shrinker *shrink,
...
+					list_add_tail(&lock->shrink_head,
+							&linfo->shrink_wlist.list);
+					spin_unlock(&linfo->shrink_wlist.lock);
+					get_lock(lock);
+					nr--;
+					freed++;
+				}
+				spin_unlock(&lock->lock);
+			}
+			put_lock(linfo, lock);
```
aversecat, line 1676: I'm seeing this `sparse` error here, on `el9` with sparse from `epel-release`:

```
  CHECK   /home/auke/scoutfs/kmod/src/lock.c
/home/auke/scoutfs/kmod/src/lock.c:1676:25: warning: context imbalance in 'lock_scan_objects' - different lock contexts for basic block
```

At first glance, to me, it seems wrong, though. It's the `put_lock` at the end of the block that triggers.

kmod/src/lock.c
```diff
@@ -1575,39 +1655,44 @@ static unsigned long lock_scan_objects(struct shrinker *shrink,
...
+	struct list_head *list;
 	unsigned long freed = 0;
 	unsigned long nr = sc->nr_to_scan;
 
 	scoutfs_inc_counter(sb, lock_scan_objects);
 
 	spin_lock(&linfo->lock);
 
-	lock = list_first_entry_or_null(&linfo->lru_list, struct scoutfs_lock, lru_head);
-	while (lock && nr > 0) {
```
zabbo: We lost this protection against entering the loop with nr == 0.  I'm not sure the modern shrinkers will ever do this,  but it doesn't look great to see that the code on its own could underflow nr and spin on all the items in the lru.  We can just move the nr == 0 test up before the first possible decrement.

kmod/src/lock.c
```diff
@@ -1495,38 +1566,47 @@ static unsigned long lock_scan_objects(struct shrinker *shrink,
...
 
-		if (nr-- == 0)
-			break;
+			spin_lock(&lock->lock);
+			if (lock->mode != SCOUTFS_LOCK_NULL && atomic_read(&lock->refcount) == 3) {
+				lock->request_pending = 1;
+				spin_lock(&linfo->shrink_wlist.lock);
+				list_add_tail(&lock->shrink_head, &linfo->shrink_wlist.list);
+				spin_unlock(&linfo->shrink_wlist.lock);
+				get_lock(lock);
```
zabbo: This is elevating the refcount for the loc's presence on the shrink list *after* giving the shrink work access to it.  It should call get_lock() up with setting request_pending before it puts it on the list.

kmod/src/lock.c
```diff
@@ -368,13 +376,15 @@ static bool insert_range_node(struct super_block *sb, struct scoutfs_lock *ins)
...
+	assert_spin_locked(&linfo->lock);
+
 	while (*node) {
 		parent = *node;
 		lock = container_of(*node, struct scoutfs_lock, range_node);
 
 		cmp = scoutfs_key_compare_ranges(&ins->start, &ins->end,
 						 &lock->start, &lock->end);
-		if (WARN_ON_ONCE(cmp == 0)) {
+		if (WARN_ON_ONCE(cmp == 0 && (atomic_read(&lock->refcount) >= 2))) {
```
zabbo, line 458: I don't think this is correct.  Once a lock overlaps with an existing lock in the range tree then its sort position in the tree is no longer trustable.  It could overlap with *everything* in the tree.  We can't continue traversing down the right child.  We could be missing an overlap in the left subtree with a lock that isn't freeing.  I think this needs to return something like the EEXIST/EINVAL like its caller.  EEXIST if it overlaps with a freeing lock so that the caller's caller will retry.  EINVAL if it overlaps with a live lock and we're bailing.

zabbo (changes requested):
First, add your Signed-off-bys for the patches you contributed to.  See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst#n409

Then just a few relatively minor things to fix.  Though minor, they're pretty subtle so that gives me some confidence that the big moving pieces are strong.

kmod/src/lock.c
chriskirby1: In main, there used to be a check for nr == 0 early in the function. Rather than even entering the loop in that case, I just put the check back.

kmod/src/lock.c
chriskirby1: Fixed.

kmod/src/lock.c
chriskirby1: Makes sense. I fixed it so we only warn on the EINVAL case and propagated the error back up the stack for a retry in the EEXIST case.

kmod/src/lock.c
chriskirby1: I'm not sure what to do with this one. I confirmed that it's complaining about our put_lock() function, not the other spinlock calls nearby. But why is it trying to say it has to be balanced within the context? Aren't there annotations that enable this checking?
Presumably we want the sparse check to be clean, so we need to work around this somehow.

zabbo: I'm not sure what it's angry about.  I'd probably try changing the code around to see if I could tease out what's making it angry.  Maybe it's warning about contexts within put_lock() when it's inlined.  Maybe it's confused by the trylock.  Nothing here seems all that fancy.  (The potential free in the put_lock is troubling, but we """"know"""" that the refcount on the list will prevent it.  Maybe it's effectively warning us that it's possible to free if the refcounts get out of whack, but that sure feels like a stretch.)

chriskirby1: I tried temporarily removing the extra call to get_lock(), but that doesn't seem to be it.

It looks like it's upset about the spin_trylock(). If I change that to a spin_lock(), sparse is happy.
If I invert the test and duplicate the call the put_lock(), sparse is happy:
```
                        if (!spin_trylock(&lock->lock)) {
                                put_lock(linfo, lock);
                        } else {
                                if (lock->mode != SCOUTFS_LOCK_NULL &&
                                    !lock->request_pending && 
                                    !lock->invalidate_pending && 
                                    atomic_read(&lock->refcount) == 3) {
                                        get_lock(lock);
                                        lock->request_pending = 1;
                                        spin_lock(&linfo->shrink_wlist.lock);
                                        list_add_tail(&lock->shrink_head,
                                                        &linfo->shrink_wlist.list);                                     
                                        spin_unlock(&linfo->shrink_wlist.lock);
                                        nr--;
                                        freed++;
                                }
                                spin_unlock(&lock->lock);
                                put_lock(linfo, lock);
                        }

```

If I then refactor out the call to put_lock(), leaving an empty if (!spin_trylock()) block, sparse becomes unhappy again.

chriskirby1: I wound up duplicating the put_lock() in each branch of the if/else.

PR #219, opened 2025-05-07
Fix test cases that don't run cleanly in a semi-dirty env.

aversecat:
All these tests don't properly clean up themselves or assume a pristine empty folder when starting, and fall over their own generated test files when repeatedly run without e.g. `-m` (mkfs each time) option passed to the test harnass. And that has implications.

I'm not a fan of tests that fully `rm -rf` everything after a test run because that makes it cumbersome to play around and inspect the test files after a test run. The most minimal change we can make is just to assure that test cases that require some clean initial state is present, will have it. This change does just that.

### DIFF
--- a/tests/run-tests.sh
+++ b/tests/run-tests.sh
@@ -532,12 +532,15 @@ for t in $tests; do
 	cmd rm -rf "$T_TMPDIR"
 	cmd mkdir -p "$T_TMPDIR"
 
-	# create a test name dir in the fs
+	# create a test name dir in the fs, clean up old data as needed
 	T_DS=""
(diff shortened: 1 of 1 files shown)


zabbo (changes requested):
I suppose we can blow away the test dirs in run-tests if it finds they exist.  We shouldn't inconsistently add rm -rf to some tests, though.

aversecat:
I'm fine with that change as well. I think we should do *both*, though.

aversecat:
I pushed a change that will erase the test directory if it already exists. The tests are no longer changed, this is just the change to the run-tests.sh script.

zabbo (approved):
Yup, looks good.

PR #220, opened 2025-05-08
Extend orphan-inodes timeout.

aversecat:
This test regularly fails in CI when the 15 seconds elapses and the system still hasn't concluded the mount log merges and orphan inode scans needed to unlink the test files.

Instead of just extending the timeout value, we test-and-retry for 120s. This hopefully is faster in most cases. My smallest VM needs about 6s-8s on average.


zabbo (approved):
Looks good.  Let's see if if any of the runners aren't covered by the generous timeout.

zabbo:
Looks good in 7/8.  95 is a steaming crater, so I guess we'll learn more when it makes it that far!

aversecat:
So, it looks like it wasn't enough. I've opened a bug for this to discuss.

PR #221, opened 2025-05-08
Give enospc test more time to commit unlink.

aversecat:
The current test sequence performs the unlink and immediately tests whether enough resources are available to create new files again, and this consistently fails.

One of my crummy VMs takes a good 12 seconds before the `touch` actually succeeds. We care about the filesystem eventually returning from ENOSPC, and certainly we don't want it to take forever, but there is a period after our first ENOSPC error and cleanup that we expect ENOSPC to fail for a bit longer.

Make the timeout 120s. As soon as the `touch` completes, exit the wait loop.

### DIFF
--- a/tests/tests/enospc.sh
+++ b/tests/tests/enospc.sh
@@ -88,6 +88,11 @@ rm -rf "$SCR/xattrs"
 
 echo "== make sure we can create again"
 file="$SCR/file-after"
+C=120
+while (( C-- )); do
+	touch $file 2> /dev/null && break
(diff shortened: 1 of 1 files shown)


zabbo:
I'm surprised this ever succeeded, honestly.  It's interesting to imagine what it took in the test hosts for log tree compaction to be fast enough to refill the metadata allocators during the removal of the xattr files.  Maybe just cores for concurrence and memory for the alloc btrees to still be resident.

PR #226, opened 2025-05-20
WIP: Prevent try_drain_data_freed() and server_log_merge_free_work() from …

chriskirby1:
…overrunning their block budget by only consuming up to 1/8 of the budget at once and calling server_apply_commit() when we hit 3/4 of the budget.


chriskirby1:
retest

PR #227, opened 2025-05-22
RHEL 9.6 support.

aversecat:
RHEL9.6 is fortunately low impact. These 2 patches come directly from my (WIP) RHEL 10 backport set, and have been through my basic testing. It looks like RH backported something that required these `mnt_idmap` patches which is to be expected as they gate a bunch of things coming with 6.x kernels.


gcymbalski:
retest

PR #233, opened 2025-06-12
Ignore sparse error about stat.h on el8.

aversecat:
On el8, sparse is at 0.6.4 in epel-release, but it fails with:
```
[SP src/util.c]
src/util.c: note: in included file (through /usr/include/sys/stat.h):
/usr/include/bits/statx.h:30:6: error: not a function <noident>
/usr/include/bits/statx.h:30:6: error: bad constant expression type
```

This is due to us needing O_DIRECT from <fcntl.h>, so we set _GNU_SOURCE before including it, but this causes (through _USE_GNU in sys/stat.h) statx.h to be included, and that has __has_include, and sparse is too dumb to understand it.

Just shut it up.

### DIFF
--- a/utils/sparse.sh
+++ b/utils/sparse.sh
@@ -22,6 +22,11 @@ RE="$RE|warning: memset with byte count of 4194304"
 # some sparse versions don't know about some builtins
 RE="$RE|error: undefined identifier '__builtin_fpclassify'"
 
+# on el8, sparse can't handle __has_include for some reason when _GNU_SOURCE
+# is defined, and we need that for O_DIRECT.
+RE="$RE|note: in included file .through /usr/include/sys/stat.h.:"
+RE="$RE|/usr/include/bits/statx.h:30:6: error: "
+
 #
 # don't filter out 'too many errors' here, it can signify that
 # sparse doesn't understand something and is throwing a *ton*


aversecat:
The last patch now just hard disables sparse during this test for the purpose of building version 1.

aversecat:
Ok, this now at least compiles OK on el8 as well and passes the format test.

aversecat:
`not ok 26 - totl-xattr-tag` :(

aversecat:
The CI failure on `debug-el8` is 
```
--- golden/offline-extent-waiting	2025-06-12 20:57:37.517111462 +0000
+++ /root/scoutfs/tests/results/output/offline-extent-waiting	2025-06-12 21:05:25.920607821 +0000
@@ -53,4 +53,5 @@
 0
 == writing waits
 should be waiting for write
+/mnt/test.0/test/offline-extent-waiting/dir/file /mnt/test.0/test/offline-extent-waiting/dir/other differ: char 688129, line 2697
 == cleanup
offline-extent-waiting output differs
```

Which is a known CI failure test, and this patch does not address.

aversecat:
The CI Failure on `debug-el95` is `INFO: task kworker/0:5:13885 blocked for more than 408 seconds.` During `large-fragmented-free`, which is a known CI failure, and this patch does not address this test.

PR #235, opened 2025-06-20
Avoid trigger munching of block_remove_stale trigger.

aversecat:
It's entirely likely that the trigger here is munched by a read on a dirty block from any unrelated or background read. Avoid that by putting the trigger at the end of the condition list.


chriskirby1:
I'll let this go through a CI run and then cherry-pick it into team/ci_green.

PR #240, opened 2025-09-11
Zab/shrink cleanup

zabbo:


zabbo:
I reworked these fixes on top of the now landed ci_green.

PR #241, opened 2025-09-12
Ignore data_version in scoutfs_ioc_data_wait_err.

aversecat:
The data_wait_err ioctl currently requires the correct data_version for the inode to be passed in, or else the ioctl returns -ESTALE. But the ioctl itself is just a passthrough mechanism for notifying data waiters, which doesn't involve the data_version at all.

Instead, we can just drop checking the value. The field remains in the headers, but we've marked it as being ignored from now on.

This all is a lot simpler than having to modify/rev the data_waiters interface to support passing back the data_version, because there isn't any space left to easily do this, and then userspace would just pass it back to the data_wait_err ioctl.


aversecat:
rebased.

PR #242, opened 2025-09-17
Check for fenced old leader in mounted test.

aversecat:
The old mounted check only considered begin/end quorum data, and not whether the old leader that is now disconnected was fenced by a new quorum leader.

Since this is the quaranteed case if the leader is disconnected forcefully, this check must account for this case, so that quorum slots can be modified if the node is permanently removed or replaced.

Normally, remounting from the fenced node should clear this. But that may be unwanted, due to e.g. equipment failure, data collection, or otherwise.

### DIFF
--- a/utils/src/util.c
+++ b/utils/src/util.c
@@ -198,11 +198,13 @@ int write_block_sync(int fd, u32 magic, __le64 fsid, u64 seq, u64 blkno,
  */
 int meta_super_in_use(int meta_fd, struct scoutfs_super_block *meta_super)
 {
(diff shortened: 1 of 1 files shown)


aversecat:
rebased

zabbo:
This needs work.   ```+funcs/exec.sh: line 60: 29153 Segmentation fault      (core dumped) "$@" >> "$T_TMP.rc.log" 2>&1```

aversecat:
hmmm double free. reworking...

aversecat:
retest

aversecat:
retest

PR #247, opened 2025-10-02
Zab/sparse error

zabbo:
This makes sparse required for userspace builds and adds it to the kernel builds.   The current ci_green commits don't add any failures.  As long as the builds succeed, I think we should merge this.  The tests may still fail without the ci_green fixes.

### DIFF
--- a/kmod/src/item.c
+++ b/kmod/src/item.c
@@ -2401,6 +2401,12 @@ out:
  * The caller has successfully committed all the dirty btree blocks that
  * contained the currently dirty items.  Clear all the dirty items and
  * pages.
+ *
+ * This strange lock/trylock loop comes from sparse issuing spurious
+ * mismatched context warnings if we do anything (like unlock and relax)
+ * in the else branch of the failed trylock.  We're jumping through
+ * hoops to not use the else but still drop and reacquire the dirty_lock
+ * if the trylock fails.
  */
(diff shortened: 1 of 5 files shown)


kmod/src/item.c
```diff
@@ -2409,41 +2415,28 @@ int scoutfs_item_write_done(struct super_block *sb)
...
-	while ((pg = list_first_entry_or_null(&cinf->dirty_list,
-					      struct cached_page,
-					      dirty_head))) {
+	do {
+		spin_lock(&cinf->dirty_lock);
+		pg = list_first_entry_or_null(&cinf->dirty_list, struct cached_page, dirty_head);
+		if (pg && write_trylock(&pg->rwlock)) {
+			list_for_each_entry_safe(item, tmp, &pg->dirty_list,
+						 dirty_head) {
+				clear_item_dirty(sb, cinf, pg, item);
```
chriskirby1, line 2424: We used to drop the cinf->dirty_lock before calling clear_item_dirty(), which also wants that lock.

zabbo: OK, I tweaked it to drop the dirty_lock around each page.  This is all so silly.

zabbo:
OK, the remaining failures are the statx warnings that we discussed fixing in ci_green.

PR #280, opened 2026-01-28
Clk/srch fixes

chriskirby1:


chriskirby1:
retest

aversecat:
retest

kmod/src/srch.c
```diff
@@ -1739,6 +1772,21 @@ static int kway_merge(struct super_block *sb,
 	/* could stream a final index.. arguably a small portion of work */
 
 out:
+	if (have_tmp) {
+		bool emit = true;
+
+		if (bl == NULL) {
+			ret = alloc_srch_block(sb, alloc, wri, sfl, &bl, blk);
```
aversecat, line 1775: I think it's possible to leak an item here if this generates ENOSPC and the conditions are right?

chriskirby1: Trying to remember once again how this works. :-)

Before starting a new input block, kway_get_reader() will  ensure we have at least 16 free blocks available. So I guess we could consume all of those in the window before we do the actual allocation, and alloc_srch_block() could return -EAGAIN. Or is there an ENOSPC possible from that path too?

aversecat: I think there's a few errors possible... -ENOMEM, -ENOSPC and -EIO in all cases from scoutfs_alloc_meta, possibly more.

chriskirby1: I think it's fine to get those errors, even with a stray lookahead item. The client will wind up setting SCOUTFS_SRCH_COMPACT_FLAG_ERROR, and the server won't use the client's compact results for that iteration.

PR #289, opened 2026-02-27
Update seq when merging deltas from partial log merge.

aversecat:
Two different clients can write delta's for totl indexes at the same time, recording their changes. When merged, a reader should apply both in order, and only once. To do so, the seq determines whether the delta has been applied already.

The code fails to update the seq while walking the trees for deltas to apply. Subsequently, when processing subsequent trees, it could re-process deltas already applied. In case of a large negative delta (e.g. removal of large amounts of files), the totl value could become negative, resulting in quota lockout.

The fix is simple: advance the seq when reading partial delta merges to avoid double counting.


zabbo (changes requested):

kmod/src/forest.c
```diff
@@ -239,6 +239,7 @@ static int forest_read_items(struct super_block *sb, struct scoutfs_key *key, u6
  * to reset their state and retry with a newer version of the btrees.
  */
 int scoutfs_forest_read_items_roots(struct super_block *sb, struct scoutfs_net_roots *roots,
+				    u64 merge_input_seq,
```
zabbo: I appreciate trying to keep the changes to a minimum, but let's not be left with a weird fragmented arg list.  In these cases make the diff noisier and reflow the args so the end result doesn't look weird.

kmod/src/forest.c
```diff
@@ -317,7 +318,15 @@ int scoutfs_forest_read_items_roots(struct super_block *sb, struct scoutfs_net_r
...
 
-		if ((le64_to_cpu(lt.flags) & SCOUTFS_LOG_TREES_FINALIZED))
+		/*
+		 * Distinguish between finalized log trees that are inputs to the
+		 * current merge and ones that are not, taking into account the merge
+		 * status item seq. scoutfs_totl_merge_contribute treats these then
+		 * as active.
+		 */
+		if ((le64_to_cpu(lt.flags) & SCOUTFS_LOG_TREES_FINALIZED) &&
+		    (!merge_input_seq ||
```
zabbo: No need to add the test for 0, nothing will be < 0.

kmod/src/forest.c
```diff
@@ -345,11 +354,61 @@ int scoutfs_forest_read_items(struct super_block *sb,
 
 	ret = scoutfs_client_get_roots(sb, &roots);
 	if (ret == 0)
-		ret = scoutfs_forest_read_items_roots(sb, &roots, key, bloom_key, start, end,
+		/* not in path of reading delta items, so we can safely pass 0 as merge
```
zabbo: Don't need a multi-line comment here, let the 0 speak for itself.  It's obvious if someone went looking, there's only a handful of callers and a single user of the FIC_ flags.

kmod/src/forest.c
```diff
@@ -345,11 +354,61 @@ int scoutfs_forest_read_items(struct super_block *sb,
...
+
+	if (super->log_merge.ref.blkno == 0) {
+		ret = 0;
+		goto out;
+	}
+
+	scoutfs_key_set_zeros(&key);
+	key.sk_zone = SCOUTFS_LOG_MERGE_STATUS_ZONE;
+
+	ret = scoutfs_btree_next(sb, &super->log_merge, &key, &iref);
```
zabbo: Just use _btree_lookup() and then you don't need the key compare when it returns 0.

kmod/src/totl.c
```diff
@@ -44,6 +44,10 @@ void scoutfs_totl_merge_contribute(struct scoutfs_totl_merging *merg,
 		merg->fin_total += le64_to_cpu(tval->total);
 		merg->fin_count += le64_to_cpu(tval->count);
 	} else {
+		/*
+		 * scoutfs_forest_read_items_roots doesn't mark non-input merges
```
zabbo: The instinct to add this comment shows that we need to back up and change some naming and comments to match the current functionality.  This is the only user of FIC_FINALIZED, so change it to MERGE_INPUT and change the fin_ prefixes in the merging struct to inp_.  With an update to the big comment above _contribute to make clear that we're talking about merge inputs, not just finalized logs, it'll all make sense.

kmod/src/wkic.c
```diff
@@ -879,9 +880,16 @@ static int insert_read_pages(struct super_block *sb, struct wkic_info *winf,
 	if (ret < 0)
 		goto out;
 
+	ret = scoutfs_forest_get_merge_input_seq(sb, &merge_input_seq);
+	if (ret == -ESTALE)
```
zabbo: Before acting on ESTALE it has to go through _check_stale so that it has a chance to recognize that it's operating on the same roots and should return hard errors for corruption rather than checking the current roots again.  A check_retry label before the call to _check_stale wouldn't be the worst.

zabbo (changes requested):

kmod/src/forest.c
aversecat: this falls away in the new version.

kmod/src/forest.c
aversecat: obsolete with the new changes.

kmod/src/totl.c
aversecat: done in new (separate) patch on top of the actual fix patch.

kmod/src/wkic.c
aversecat: obsolete, no longer needed with the seq from the client log merge.

kmod/src/forest.c
aversecat: obsolete

kmod/src/forest.c
aversecat: If this is 0 then there is no log merge in progress, it was just completed and we have to treat any finalized tree as already merged with fs_root - they just need to be freed. So we tag them as FIC_MERGE_ITEM so the `inp_seq > fs_seq check` check in _resolve properly dispatches them. So this part needs to remain. See the new version of the patch.

aversecat:
Hmm, the one failure is the xfs lockdep (I didn't rebase this yet, so it hit it) but the other failure looks bad, so I'll wait for the vmcore tomorrow morning to see what it is about.

kmod/src/wkic.c
```diff
@@ -805,33 +806,72 @@ static void free_page_list(struct super_block *sb, struct list_head *list)
...
+	struct scoutfs_log_merge_status *stat;
+	SCOUTFS_BTREE_ITEM_REF(iref);
+	struct scoutfs_key key;
 	int ret;
 
 	mutex_lock(&winf->roots_mutex);
 
 	if (force_new || ktime_before(winf->roots_expire, ktime_get_raw())) {
-		ret = scoutfs_client_get_roots(sb, &roots);
+		super = kmalloc(sizeof(*super), GFP_NOFS);
```
zabbo: Extract this whole thing into its own function.  It keeps all the stuff it needs out of the fast path function that uses the current roots that haven't expired.

kmod/src/wkic.c
```diff
@@ -805,33 +806,72 @@ static void free_page_list(struct super_block *sb, struct list_head *list)
...
+			.srch_root = super->srch_root,
+		};
+
+		winf->merge_input_seq = 0;
+		if (super->log_merge.ref.blkno) {
+			scoutfs_key_set_zeros(&key);
+			key.sk_zone = SCOUTFS_LOG_MERGE_STATUS_ZONE;
+			ret = scoutfs_btree_lookup(sb, &super->log_merge, &key, &iref);
+			if (ret == 0) {
+				if (iref.val_len == sizeof(*stat)) {
```
zabbo: Return -EUCLEAN if the item wasn't the right size.   We don't want to silently proceed in the face of .. I don't know, weird version incompatibility bugs or corruption.

zabbo (changes requested):
Just a few easy cleanups and we're good!  I think!

kmod/src/wkic.c
aversecat: :alien:

aversecat:
Looks like one job is stuck on our unmount issue, unrelated to this PR.

aversecat:
retest

aversecat:
1. el96 has `archive-light-cycle unexpected messages in dmesg` but there's no actual dmesg output in the `fail.log`, however, the serial console has:
```
[ 6676.234424] 
[ 6676.234432] ======================================================
[ 6676.234433] WARNING: possible circular locking dependency detected
[ 6676.234437] 5.14.0-611.35.1.el9_7.x86_64+debug #1 Tainted: G           OE      ------  --- 
[ 6676.234439] ------------------------------------------------------
[ 6676.234440] kswapd0/119 is trying to acquire lock:
[ 6676.234443] ff1100015be38120 (&xfs_nondir_ilock_class){++++}-{4:4}, at: xfs_can_free_eofblocks+0x349/0x540 [xfs]
```

edit: this is the XFS lockdep we're trying to squelch....

So despite rebase this is somehow trickling through and failing the test.

2. el97 aborted run right after initial mount.

aversecat:
retest

kmod/src/btree.c
```diff
@@ -2486,6 +2492,16 @@ int scoutfs_btree_merge(struct super_block *sb,
 			mitem = next_mitem(mitem);
 			free_mitem(&rng, tmp);
 		}
+
+		if (nr_modified >= le16_to_cpu(bt->nr_items) &&
```
zabbo: We don't need to track nr_modified.  We want to return erange when we've finished with a block without the weird split/merge case, and that there's still items to merge.  And we don't want to burn the trigger if we're not going to return, which this will do when the items are exhausted.  lose nr_modified and just have "if (mitem && walk_val_len == 0 && !(walk_flags & (BTW_INSERT|BTW_DELETE) && _trigger) and then the usual ret = and next_ret =.  I think that'll do it.

zabbo (changes requested):

kmod/src/btree.c
aversecat: applied.

aversecat:
Once more it hit:
```
[ 3268.610027] ======================================================
[ 3268.610029] WARNING: possible circular locking dependency detected
[ 3268.610031] 5.14.0-611.35.1.el9_7.x86_64+debug #1 Tainted: G           OE      ------  --- 
[ 3268.610034] ------------------------------------------------------
[ 3268.610035] kswapd0/111 is trying to acquire lock:
[ 3268.610038] ff1100011b09a220 (&xfs_nondir_ilock_class){++++}-{4:4}, at: xfs_can_free_eofblocks+0x349/0x540 [xfs]
```

aversecat:
retest

PR #298, opened 2026-04-13
Fix double unlock in scoutfs_setattr data_wait error path

aversecat:
When scoutfs_setattr truncates a file with offline extents, it unlocks the inode lock before calling scoutfs_data_wait to wait for the data to be staged. If data_wait returns any error, the code jumps to 'goto out' which calls scoutfs_unlock again, thus double-unlocking the lock.


aversecat:
retest

aversecat:
rebased

PR #299, opened 2026-04-13
Add cond_resched in block_free_work

aversecat:
I'm seeing consistent CPU soft lockups in block_free_work on my bare metal system that aren't reached by VM instances. The reason is that the bare metal machine has a ton more memory available causing the block free work queue to grow much larger in size, and then it has so much work that it can take 30+ seconds before it goes through it all.

This is all with a debug kernel. A non debug kernel will likely zoom through the outstanding work here at a much faster rate.


aversecat:
retest

PR #305, opened 2026-04-20
Set BLOCK_BIT_ERROR on bio submit failure during forced unmount

aversecat:
block_submit_bio will return -ENOLINK if called during a forced shutdown, the bio is never submitted, and thus no completion callback will fire to set BLOCK_BIT_ERROR. Any other task waiting for this specific bp will end up waiting forever.

To fix, when -ENOLINK is returned, set it here ourselves so the other waiters drain.

It's possible that nothing else wakes up the waiters at this point since we're in forced shutdown and it might very well be the last bio, so we manually wake up the remainder to avoid a stall.

### DIFF
--- a/kmod/src/block.c
+++ b/kmod/src/block.c
@@ -467,9 +467,6 @@ static int block_submit_bio(struct super_block *sb, struct block_private *bp,
 	sector_t sector;
 	int ret = 0;
 
-	if (scoutfs_forcing_unmount(sb))
-		return -ENOLINK;
-
 	sector = bp->bl.blkno << (SCOUTFS_BLOCK_LG_SHIFT - 9);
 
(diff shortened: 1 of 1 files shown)


zabbo (changes requested):
Nice find!  But instead of bringing up part of the end_io code into callers, change block_submit_bio() to use its error path that calls end_io with the error instead of just returning.  That'd also catch the case over in the write side.  (It won't hang, but it will return 0 for writes that were never submitted).

aversecat:
OK, I had looked into the write side as well and didn't see anything, but, this will be cleaner for sure.

zabbo (approved):

aversecat:
CI didn't start any of the jobs except 2, so I've started a second run.

PR #307, opened 2026-04-29
Search all merge range items for next

zabbo:
When searching for the next least merge range we need to sweep all the stored items because they're interleaved with respect to key sorting because we've clobbered the zone.

To search all of them we need to start from 0, not from the caller's start key after setting the zone.  If the caller happens to provide a start key with a small zone but large other fields (totl keys with sufficiently large identifiers) we can miss ranges.

### DIFF
--- a/kmod/src/server.c
+++ b/kmod/src/server.c
@@ -1077,8 +1077,7 @@ static int next_log_merge_range(struct super_block *sb, struct scoutfs_btree_roo
 	struct scoutfs_key key;
 	int ret;
 
-	key = *start;
-	key.sk_zone = SCOUTFS_LOG_MERGE_RANGE_ZONE;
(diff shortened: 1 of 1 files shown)


kmod/src/server.c
```diff
@@ -1077,7 +1077,7 @@ static int next_log_merge_range(struct super_block *sb, struct scoutfs_btree_roo
 	struct scoutfs_key key;
 	int ret;
 
-	key = *start;
+	scoutfs_key_set_zeros(&key);
 	key.sk_zone = SCOUTFS_LOG_MERGE_RANGE_ZONE;
```
aversecat: init_log_merge_key() ?

zabbo: Sure!

aversecat (approved):
lgtm

PR #309, opened 2026-05-01
fix and test - quota invalidate race

aversecat:
This was stashed on my branch with stress testing related things I had run into a few weeks or so back. The race that exposes the BUG_ON here is not very complicated: quota readers race a single writer (quota mod rule). Fix and test case.

### DIFF
--- a/kmod/src/quota.c
+++ b/kmod/src/quota.c
@@ -1114,6 +1114,7 @@ int scoutfs_quota_mod_rule(struct super_block *sb, bool is_add,
 			goto release;
 	}
 
+	wait_event(qtinf->waitq, !ruleset_is_busy(qtinf));
 	scoutfs_quota_invalidate(sb);
 	ret = 0;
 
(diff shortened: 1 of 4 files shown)


kmod/src/quota.c
```diff
@@ -730,7 +730,14 @@ static int read_ruleset(struct super_block *sb, struct squota_info *qtinf)
 		}
 
 		spin_lock(&qtinf->lock);
-		rcu_assign_pointer(qtinf->ruleset, rs);
+		/* drop our result if a racing invalidation cleared EBUSY */
```
zabbo: The point of setting EBUSY is to only have one reader of the ruleset on the node at a time.  If we were going to let it be dropped, we probably just wouldn't use it at all.  It'd make the code simpler.  But I don't think we want invalidation to clear busy.  So this change can be dropped.

kmod/src/quota.c
zabbo: Change this so that it assigns on == ENOENT || !IS_ERR.  So it'll do nothing if it's busy.  That's fine for the lock invalidation case, and we'll make it fine for the mod case.  (And the nice comment update can be tweaked a bit to reflect this.)

zabbo, line 1118: The only problem here is that a pending read_rules can be assembling the rules before the change and can assign them after the invalidate.  If we add a wait_event for it to be !busy before the invalidate then we're sure that the new rules will be used before we return.

zabbo (changes requested):
Great find, and the fix is close.  We can just tweak it a bit so we don't replace ebusy in invalidation and instead have mod wait for a pending reader to finish before it invalidates.

kmod/src/quota.c
aversecat: Thanks, that does turn out to be a lot cleaner and much better to follow.

aversecat:
retest

aversecat:
two failed forced unmount CI failures. Unrelated.

aversecat:
one more unmount CI failure.

aversecat:
retest

aversecat:
still another CI unmount failure, but nothing else.

zabbo (approved):
Lookin' good!

PR #311, opened 2026-05-07
Optional per-kver kmod package builds

gcymbalski:
This allows optional-at-build-time per-kver kmod packages by defining `per_kver` to `1` at build time, which means:
- Package names use conventions used elsewhere in RHEL to indicate the kernel version built against
- A virtual package, `kmod-scoutfs`, preserves the existing name and also makes it simple to install with yum via repository
- Strict requirements on the kver built against in accordance with conventions
- Disabling automatic use of `weak-modules`

If *not* defining `per_kver` to `1`, legacy behavior is preserved, meaning we can deliver packages for the same scoutfs commit in both styles depending on customer needs.


aversecat (approved):
lgtm

PR #314, opened 2026-05-12
Drop pending greetings from resend_queue in set_valid_greeting.

aversecat:
Also some trace events used to debug this - high value for recovery troubleshooting ones.


aversecat:
The spin_lock_bh change was precursor to this patch, and not independent. It needs to precede the trace event introduction. I'm not sure this is worth the hassle - we could just drop the trace events and the spin_lock_bh change from this patch, and go with only the greeting drop patch here.

PR #315, opened 2026-05-12
Clear ref_blkno output when block is already dirty

aversecat:
block_dirty_ref() skipped setting *ref_blkno when the block was already dirty, leaving the caller with a stale value passed by reference.

dirty_alloc_blocks() calls it twice but when the referenced block is already dirty, it will receive the uninitialized stack value back, and then adding it freed list with list_block_add() later.

Set it to 0 on the already-dirty fast path so callers do not try to free a random block.

### DIFF
--- a/kmod/src/block.c
+++ b/kmod/src/block.c
@@ -836,6 +836,9 @@ int scoutfs_block_dirty_ref(struct super_block *sb, struct scoutfs_alloc *alloc,
 	int ret;
 	int err;
 
+	if (ref_blkno)
+		*ref_blkno = 0;
+
 	/* read existing referenced block, if any */
 	blkno = le64_to_cpu(ref->blkno);
 	if (blkno) {


kmod/src/block.c
```diff
@@ -846,6 +846,8 @@ int scoutfs_block_dirty_ref(struct super_block *sb, struct scoutfs_alloc *alloc,
 		bp = BLOCK_PRIVATE(bl);
 
 		if (block_is_dirty(bp)) {
+			if (ref_blkno)
```
zabbo: It'd be better to zero it at the start of the function so that all return paths get the initialized zero.  The commit message is also misleading.  It won't ever be called on a dirty block because dirty_alloc_blocks bails when the blocks are already dirty.  We still want to fix it but it's harmless with current callers.

zabbo (changes requested):

kmod/src/block.c
aversecat: gotcha, I've moved it up high and addressed the commit msg gaps.

PR #316, opened 2026-05-14
el7 support removal

aversecat:
I've had a bunch of these already prepped for a while now, most of it is just `unifdef` work. Let it cook in CI for a while before we look at it later this year.


aversecat:
reminder I'll need to update the spec files that have `%{?el7}` macros in them....

aversecat:
retest

aversecat:
This should be ready - obviously CI will fail on the el7 builds. All others are green, though!

PR #323, opened 2026-06-18
Auke/freed list accounting

aversecat:
This is a cleaned up branch resolving the issues we encountered over the past week and helped unwedge a stuck commit.


utils/src/print.c
```diff
@@ -419,8 +419,17 @@ static int print_log_merge_item(struct scoutfs_key *key, u64 seq, u8 flags, void
 	struct scoutfs_log_merge_complete *comp;
 	struct scoutfs_log_merge_freeing *fr;
 
+	/*
+	 * Parent block items reference child blocks and have no value;
+	 * print_block_ref() calls us with a NULL val just to print the key.
+	 */
+	if (!val)
```
zabbo, line 452: Nice catch.  Might as well fix the srch_root and mounted_client prints while you're in here.  They seem to have the same problem.

kmod/src/fence.c
```diff
@@ -238,6 +238,25 @@ int scoutfs_fence_start(struct super_block *sb, u64 rid, __be32 ipv4_addr, int r
...
+	 * shutting the server down.
+	 */
+	spin_lock(&fi->lock);
+	list_for_each_entry(fence, &fi->list, entry) {
+		if (fence->rid == rid) {
+			spin_unlock(&fi->lock);
+			return 0;
+		}
+	}
+	spin_unlock(&fi->lock);
```
zabbo: Let's rework this to not add a mid function return and to have it put the fence on the list.  Could use a bool to record that it found the rid on the list and goto out after unlocking.  Otherwise it put it on the list and is then reserving the rid as it goes to create the sysfs files.  If that fails the server is shutting down so we don't really care what other paths think of the fencing state if we fail to add the sysfs files.  The exit path can then try and free the fence if it's non-null and also try and remove it from the list before freeing.  The success path can set the fence to null so it doesn't try to remove.

kmod/src/fence.c
```diff
@@ -258,6 +277,9 @@ int scoutfs_fence_start(struct super_block *sb, u64 rid, __be32 ipv4_addr, int r
 						&fence->ssa, fence_attrs,
 						"%016llx", rid);
 	if (ret < 0) {
+		/* highly unlikely race collision */
+		if (ret == -EEXIST)
```
zabbo: Let's not  mask this error.  If we get this then sysfs has already sent an error message to the console.  We want to avoid this case entirely.

zabbo (changes requested):
Just two little fixes and this all looks good.

utils/src/print.c
aversecat: Yep, spotted them, same fix there.

kmod/src/fence.c
aversecat: Thanks, this makes sense now. Yes, that avoids the need for the error masking. I'll redo.

aversecat:
retest

aversecat:
@zabbo rebased and now has the test at the top of the stack.

zabbo:
Hmm, did the rebase miss some changes?  The fencing fix is still only testing under the first lock, not adding it to reserve the rid, and still has the error masking.  And the userspace print changes only touches the one function?  But with those fixed it's fine.

aversecat:
> Hmm, did the rebase miss some changes? The fencing fix is still only testing under the first lock, not adding it to reserve the rid, and still has the error masking. And the userspace print changes only touches the one function? But with those fixed it's fine.

argh I must have pushed the test from a stale branch.

aversecat:
OK, branch restored to `783fdcb` with the test case cherry picked on top of it.

kmod/src/fence.c
```diff
@@ -253,23 +256,54 @@ int scoutfs_fence_start(struct super_block *sb, u64 rid, __be32 ipv4_addr, int r
 	fence->error = false;
 	fence->reason = reason;
 	fence->rid = rid;
+	/* init the timer before list_add so a racing fence_stop can del_timer_sync it */
```
zabbo: There had better not be racing fence_stop with _start.  If there was, it could find the fence struct on the list and free it out from under us.  The server is careful to stop callers of fencing (reclaim and recov work) before calling fence_stop.  Moving the init up here is a good fix, so it's by the other init of a complicated struct member, but just remove the (hopefully!) incorrect comment.  This does raise the real problem that fence structs can be freed once they're on the list, though.  The whole life cycle is built around that.  Rather than putting it on the list early, I think we start with the old code and 1) add a check for an existing fence on the list before creating the sysfs files 2) wrap a fi->mutex around the whole check_list/add_sysfs/add_list block, and 3) grab that same mutex around removing the fence from the list and removing the sysfs files.  That'll take some refactoring of the freeing path.  (which you can call from there.. initing the timer/attrs lets us shut them down regardless of if they were added/created or not).

zabbo (changes requested):
Looks like we need to redo the fence add/remove safety a bit, but it's close.

kmod/src/fence.c
aversecat: Thanks, that didn't turn out that complex I think. That mutex seems entirely reasonable. Letting the robots try it out now.

zabbo (approved):

PR #328, opened 2026-07-11
Use raft pre-vote to avoid election term inflation.

aversecat:
A quorum member that loses contact with the majority used to increment and persist its term on every election timeout while it could never win. A network-partitioned former leader could climb its term far above the cluster's, then on reboot read that inflated term back from its quorum block and usurp the leader that legitimately replaced it.

The previous term-persist approach (deferring the durable term write until a vote arrived) did not cover a partition that splits off an intercommunicating sub-majority: in a five node cluster a two node minority that can still reach each other exchange votes, so both the voter (adopting the higher term) and the candidate (seeing a vote) persist the climbing term anyway.

Add a raft pre-vote round.  On election timeout a member becomes a pre-candidate and probes for support at its prospective next term without incrementing or persisting it.  Peers grant a pre-vote only when they do not currently see a live leader, and pre-vote messages are excluded from the adopt-greater-term path so granting changes no state. Only once a majority of pre-votes is collected does the member increment its term, persist it, and stand for election for real.  A partitioned member or sub-majority can never reach a majority of pre-votes, so it loops pre-vote rounds at a frozen term, never inflating it, and follows the majority's leader once the partition heals.

### DIFF
--- a/kmod/src/format.h
+++ b/kmod/src/format.h
@@ -763,7 +763,11 @@ struct scoutfs_quorum_message {
 #define SCOUTFS_QUORUM_MSG_HEARTBEAT	2
 /* leaders broadcast as they leave to break heartbeat timeout */
 #define SCOUTFS_QUORUM_MSG_RESIGNATION	3
-#define SCOUTFS_QUORUM_MSG_INVALID	4
+/* a pre-candidate probes for support at its prospective next term */
+#define SCOUTFS_QUORUM_MSG_REQUEST_PREVOTE	4
+/* peers grant pre-votes when they don't see a live leader */
+#define SCOUTFS_QUORUM_MSG_PREVOTE	5
+#define SCOUTFS_QUORUM_MSG_INVALID	6
 
 /*
  * The version is currently always 0, but will be used by mounts to
--- a/kmod/src/quorum.c
+++ b/kmod/src/quorum.c
@@ -69,6 +69,24 @@
  * backwards in time so each mount also uses its quorum block to store
  * the greatest term it has used in messages.
  *
+ * Before a member increases and persists its term to stand for election
+ * it first runs a raft pre-vote round: as a pre-candidate it asks the
+ * others to support its prospective next term without adopting that term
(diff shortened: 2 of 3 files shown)


kmod/src/format.h
```diff
@@ -763,7 +763,11 @@ struct scoutfs_quorum_message {
 #define SCOUTFS_QUORUM_MSG_HEARTBEAT	2
 /* leaders broadcast as they leave to break heartbeat timeout */
 #define SCOUTFS_QUORUM_MSG_RESIGNATION	3
-#define SCOUTFS_QUORUM_MSG_INVALID	4
+/* a pre-candidate probes for support at its prospective next term */
+#define SCOUTFS_QUORUM_MSG_REQUEST_PREVOTE	4
+/* peers grant pre-votes when they don't see a live leader */
+#define SCOUTFS_QUORUM_MSG_PREVOTE	5
+#define SCOUTFS_QUORUM_MSG_INVALID	6
```
zabbo, line 770: This changes the network format and should be protected by a format version change.  If we had tests that checked forward/back compat they'd have failed.  That alone makes me think this isn't the right approach.

kmod/src/quorum.c
```diff
@@ -804,10 +880,46 @@ static void scoutfs_quorum_worker(struct work_struct *work)
...
 		}
 
-		/* followers and candidates start new election on timeout */
+		/*
+		 * Grant a pre-vote for a higher prospective term as long as
+		 * we don't currently see a live leader.  Pre-votes don't
+		 * change our term, role, or recorded vote so we can grant
+		 * them freely.  Refusing while a leader is heartbeating keeps
+		 * a spuriously timed-out member from disrupting a healthy
+		 * leader.
```
zabbo, line 893: I'm not convinced that this is safe.  One agent hearing heartbeats from a leader doesn't mean everyone is, and it doesn't mean that the leader is healthy and shouldn't be disrupted.

zabbo (changes requested):
This is all too risky.  We need to rethink the approach here.

PR #331, opened 2026-07-30
Fix warning about missing cluster lock during readahead

chriskirby1:
scoutfs_readahead() will acquire the cluster read lock if the task doesn't already hold it. But it doesn't set the corresponding per-task variable, so the scoutfs_get_block() call won't find it and mistakenly warns that we aren't holding the lock.

We need to set the per-task variable in scoutfs_readahead(). And since we can get into this path from scoutfs_fadvise(), and that function acquires the extent_sem, we also need to first acquire the cluster read lock and set the per-task variable there too. This fixes a lock ordering issue that we had not yet seen.

### DIFF
--- a/kmod/src/data.c
+++ b/kmod/src/data.c
@@ -779,6 +779,7 @@ static void scoutfs_readahead(struct readahead_control *rac)
 {
 	struct inode *inode = rac->file->f_inode;
 	struct scoutfs_inode_info *si = SCOUTFS_I(inode);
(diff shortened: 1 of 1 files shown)


chriskirby1:
lock-recover-invalidate, oy. retest

zabbo (approved):

aversecat (approved):
lgtm, no longer throws the WARN_ON, even after a fresh boot.

